PR #20, opened 2024-10-29
bug: Add failing test for bug caused by casting kernel before launch

johnbowen42:


davidbeckingsale:
The `indirect launcher` test doesn't fail, but we could add this case to that file. Is that what you are suggesting?

ggeorgakoudis:
It doesn't fail because the lit CHECK expressions are crafted to assume that Proteus will not handle indirect launching. We should fix indirect launching and change CHECK expressions to expect 2 (non-cached) Proteus JIT executions for the first run and 2 hits in the storage cache for the second run.

davidbeckingsale:
Got it, thanks for the clarification. I think it would be good to mark expected failure cases more clearly in the tests.

ggeorgakoudis:
I agree, I haven't found a good way to do it with ctest. Marking `WILL_FAIL` will pass the tests when it fails, with no indication in the output that this is an expected failure. Ideally, we would like to test suite to pass but report that some tests fail. Is there a way to do that?

ggeorgakoudis:
Superseded by #36

PR #33, opened 2024-11-12
feature: Support indirect kernel launching via static analysis

johnbowen42:
Instead of looking directly at kernel launch sites, BFS through the def-use chains of each function annotated with JIT.  When a CallBase is reached, continue through the def use chains of the Instructions inside the called function recursively, until a kernel launch site is reached


ggeorgakoudis:
@johnbowen42 Can we close this? It's far from our current development plan. We can open an internal discussion to store the idea.

johnbowen42:
> @johnbowen42 Can we close this? It's far from our current development plan. We can open an internal discussion to store the idea.

Definitely, let's close for now.  This is not currently useful because indirect launching is well supported.

PR #38, opened 2024-11-26
Support newer versions of LLVM and ROCm

tbennun:
Adds support for newer versions of LLVM (18) and ROCm (6.2).

### DIFF
--- a/lib/JitEngine.cpp
+++ b/lib/JitEngine.cpp
@@ -21,7 +21,6 @@
 #include "llvm/IRReader/IRReader.h"
 #include "llvm/MC/TargetRegistry.h"
 #include "llvm/Passes/PassBuilder.h"
-#include "llvm/Support/CodeGen.h"
 #include "llvm/Target/TargetMachine.h"
 #include "llvm/TargetParser/SubtargetFeature.h"
 #include <llvm/IR/Constants.h>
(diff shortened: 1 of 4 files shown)


lib/JitEngine.cpp
```diff
@@ -42,7 +43,11 @@ using namespace llvm;
 Expected<std::unique_ptr<TargetMachine>>
 JitEngine::createTargetMachine(Module &M, StringRef Arch, unsigned OptLevel) {
   Triple TT(M.getTargetTriple());
+#if LLVM_VERSION_MAJOR >= 18
+  std::optional<llvm::CodeGenOptLevel> CGOptLevel = CodeGenOpt::getLevel(OptLevel);
```
tbennun, line 47: One could also just use `auto` here

ggeorgakoudis: Let's go for `auto` to avoid preprocessor pollution

tbennun: Done.

tbennun:
> We need to have tests ran for LLVM 18 too. However, github and tioga tests succeeded on LLVM 17, which makes me believe that your changes are portable for both LLVM 17 and 18?

The functions I'm using (`PointerType::get`) are compatible going back to LLVM 14 (and perhaps even earlier).

I also addressed your comment and it still works for me.

ggeorgakoudis:
Supereded #44

PR #39, opened 2024-11-26
Add a test (expected to fail) for calls to device function in different translation units

koparasy:
When calling a proteus annotated kernel that calls an external device function with rdc proteus fails.

AMD:

```
$ ./tests/gpu/device_function.HIP.rdc
ERROR @ /usr/workspace/koparasy/proteus/lib/JitEngineDeviceHIP.cpp:169 ->  HIPRTC_ERROR_LINKING
Aborted (core dumped)
```

CUDA:

```

ERROR @ /usr/workspace/koparasy/proteus/lib/JitEngineDeviceCUDA.cpp:57 ->  device kernel image is invalid
Aborted
```


tests/gpu/CMakeLists.txt
```diff
@@ -73,7 +73,14 @@ function(CREATE_GPU_TEST_RDC exe source)
...
 
-    add_test(NAME ${exe}.${lang}.rdc COMMAND ${LIT} -vv -D EXT=${lang}.rdc ${source})
+    # Code to pull only lit instructions from first source file
+    list(LENGTH source num_source_files)
+    if(${num_source_files} GREATER 1)
+      list(GET source 0 lit_instructions)
+      add_test(NAME ${exe}.${lang}.rdc COMMAND ${LIT} -vv -D EXT=${lang}.rdc ${lit_instructions})
+    else()
+      add_test(NAME ${exe}.${lang}.rdc COMMAND ${LIT} -vv -D EXT=${lang}.rdc ${source})
+    endif()
```
ggeorgakoudis, line 83: Do we need this? We already have a multi-file test (`file1_kernel.cpp`, `file2_kernel.cpp`) and lit picks up checks from the first file

tests/gpu/CMakeLists.txt
```diff
@@ -133,6 +140,14 @@ CREATE_GPU_TEST_RDC(kernel_host_jit kernel_host_jit.cpp)
...
 CREATE_GPU_TEST_RDC(types types.cpp)
 
+# The following tests fails and need to be "fixed" I pass will fail true to enable CI
+CREATE_GPU_TEST_RDC(device_function "main_code.cpp;device_code.cpp")
+if(ENABLE_CUDA)
+set_tests_properties(device_function.CUDA.rdc PROPERTIES WILL_FAIL TRUE)
+else()
+set_tests_properties(device_function.HIP.rdc PROPERTIES WILL_FAIL TRUE)
+endif()
+
```
ggeorgakoudis, line 150: Add a `will_fail` flag in the `CREATE_GPU_TEST_RDC` function to set it there and avoid if-else

tests/gpu/main_code.cpp
```diff
@@ -0,0 +1,37 @@
+// clang-format off
+// RUN: ./rdc.%ext | FileCheck %s --check-prefixes=CHECK,CHECK-FIRST
+// Second run uses the object cache.
+// RUN: ./rdc.%ext | FileCheck %s --check-prefixes=CHECK,CHECK-SECOND
```
ggeorgakoudis, line 4: Naming for the run program seems off. It should be the executable plus the lit-provided extensions. I guess `main_code.%ext`

tests/gpu/main_code.cpp
ggeorgakoudis, line 1: Rename `kernel_calls_func.cpp`

tests/gpu/device_code.cpp
ggeorgakoudis, line 1: Rename `device_func.cpp`

tests/gpu/CMakeLists.txt
ggeorgakoudis: Does it break only for RDC compilation? Shouldn't it break for non-RDC too?

tests/gpu/CMakeLists.txt
koparasy: On the rdc tests you do not have multi-file tests and are not taking into account the ARGN function parameter.

Now that both functions have an extra argument (will_fail) I bundle all sources together as a list on the second argument.

koparasy:
@ggeorgakoudis I addressed your comments and I added on more test.

The additional test generates a library from device code (when building the library the pass is turned off). The test links in with that code and executes it. This currently fails because of the original rdc issue we have. Once this [branch](https://github.com/Olympus-HPC/proteus/tree/support-rdc-device-funcs) lands I believe the test will manifest one more issue we need to fix.

PR #42, opened 2024-12-03
Specializes module based on BlockDim.* values and GridDim.* values.

koparasy:
@ggeorgakoudis @tomstitt this is what we discussed on the meeting for CUDA. We specialize the values and we inform llvm about the range of value of `threadIdx.*`. I have not checked whether this improves performance. Any ideas on how I can test this are welcome. The todos are:

- [ ] Implement the AMD
- [ ] Run the CGO artifact branch with these changes and compare performance.
- [ ] @ggeorgakoudis currently a single env flag enables both optimizations (assume && specialization). Is this what you want?


koparasy:
I don't think this will remove the `for` loop. I believe LLVM is unaware of the SPMD paradigm of CUDA/HIP. So, it cannot prove that this function will be spawned K times. But this is a step towards the right direction.

koparasy:
A summary of the results:

Main branch:

main

Here are the results of this branch:

new_opt

## NVIDIA

Reporting speedup in respect to AOT for the main branch:

figure-3-bar-end2end-speedup-nvidia

versus the new optimization:

figure-3-bar-end2end-speedup-nvidia

and for NVIDIA here are the plots when looking only at the kernel execution time:

Main Branch:

figure-4-bar-kernel-speedup-nvidia

This Branch:

figure-4-bar-kernel-speedup-nvidia


In summary, for NVIDIA it doesn't seem to have big gains, however it does not inflate execution time etc. 

## AMD

Reporting speedup in respect to AOT for the main branch:

figure-3-bar-end2end-speedup-amd

This branch:

figure-3-bar-end2end-speedup-amd


Reporting speedup by taking into account only kernel execution time:

Main Branch:

figure-4-bar-kernel-speedup-amd

This branch:

figure-4-bar-kernel-speedup-amd

Overall on AMD there seem to be gains. The bar plots do not include LULESH, yet in AMD we drop the total execution time by 5 seconds.

koparasy:
I also checked this attributes: https://llvm.org/docs/AMDGPUUsage.html#id132 apparently AMD can auto-detect the threadIdx/blockIdx invariant code and optimize accordingly.

lib/JitEngineDeviceHIP.cpp
```diff
@@ -132,6 +133,79 @@ void JitEngineDeviceHIP::setLaunchBoundsForKernel(Module &M, Function &F,
              << WavesPerEU << "\n");
 }
 
+void JitEngineDeviceHIP::setKernelDims(Module &M, dim3 &GridDim,
```
ggeorgakoudis, line 136: There is a lot of code duplication between the HIP and CUDA cases. Move `setKernelDims` to `JitEngineDevice.hpp`. Create CRTP specialized getters for dimension functions that return the instrinsic/function name.

ggeorgakoudis (changes requested):
This is great, and the performance report in the PR is excellent. Requesting refactoring changes to avoid code duplication and other minor issues.

lib/JitEngineDeviceCUDA.cpp
```diff
@@ -97,6 +97,65 @@ void JitEngineDeviceCUDA::setLaunchBoundsForKernel(Module &M, Function &F,
...
+    }
+  };
+
+  InsertAssume("llvm.nvvm.read.ptx.sreg.tid.x", BlockDim.x);
+  InsertAssume("llvm.nvvm.read.ptx.sreg.tid.y", BlockDim.y);
+  InsertAssume("llvm.nvvm.read.ptx.sreg.tid.z", BlockDim.z);
+
+  InsertAssume("llvm.nvvm.read.ptx.sreg.ctaid.x", BlockDim.x);
+  InsertAssume("llvm.nvvm.read.ptx.sreg.ctaid.y", BlockDim.y);
+  InsertAssume("llvm.nvvm.read.ptx.sreg.ctaid.z", BlockDim.z);
```
koparasy, line 156: This part of the code is wrong. It should be GridDim.* . I will fix it the commit.

lib/JitEngineDeviceCUDA.cpp
```diff
@@ -97,6 +97,65 @@ void JitEngineDeviceCUDA::setLaunchBoundsForKernel(Module &M, Function &F,
...
+                                        dim3 &BlockDim) {
+  auto ReplaceIntrinsicDim = [&](StringRef IntrinsicName, uint32_t value) {
+    Function *IntrinsicFunction = M.getFunction(IntrinsicName);
+    if (!IntrinsicFunction)
+      return;
+
+    for (auto U = IntrinsicFunction->use_begin(),
+              UE = IntrinsicFunction->use_end();
+         U != UE;) {
+      Use &Use = *U++;
```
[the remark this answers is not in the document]
koparasy: I opted to not use the `auto` approach for safety reasons.

lib/JitEngineDeviceCUDA.cpp
koparasy: I executed once more manually wsm5 with the fix and observed a similar speedup. I assume the perf numbers I send earlier are still correct.

ggeorgakoudis (approved):

PR #44, opened 2024-12-03
Fix for LLVM 18 compatibility and CI

ggeorgakoudis:
- Changes for compatibility with LLVM 18 and LLVM 17
- Cleanup cmake files
- CI for LLVM 18 through rocm 6.2.1

---------


pass/ProteusPass.cpp
```diff
@@ -497,19 +572,10 @@ struct ProteusJitPassImpl {
...
-
-    FunctionType *JitEntryFnTy =
-        FunctionType::get(VoidPtrTy,
-                          {VoidPtrTy, VoidPtrTy, Int32Ty,
-                           RuntimeConstantTy->getPointerTo(), Int32Ty},
-                          /* isVarArg=*/false);
+    FunctionType *JitEntryFnTy = FunctionType::get(
+        PtrTy,
+        {PtrTy, PtrTy, Int32Ty, RuntimeConstantTy->getPointerTo(), Int32Ty},
+        /* isVarArg=*/false);
```
koparasy, line 578: Create a function in `ProteusJitPassImpl` (`getOrInsertProteusEntry`) that creates both the FnTy and the FunctionCallee.

pass/ProteusPass.cpp
```diff
@@ -695,25 +756,24 @@ struct ProteusJitPassImpl {
...
-        FunctionType::get(Int32Ty,
-                          {ModuleUniqueId->getType(), VoidPtrTy,
-                           FatbinWrapper->getType(), Int64Ty, Int64Ty, Int32Ty,
-                           Int64Ty, Int32Ty, VoidPtrTy, Int64Ty, VoidPtrTy},
-                          /* isVarArg=*/false);
+    JitEntryFnTy = FunctionType::get(
+        Int32Ty,
+        {ModuleUniqueId->getType(), PtrTy, FatbinWrapper->getType(), Int64Ty,
+         Int64Ty, Int32Ty, Int64Ty, Int32Ty, PtrTy, Int64Ty, PtrTy},
+        /* isVarArg=*/false);
```
koparasy, line 763: I was confused by the naming convention: `JitEntryFnTy` maps to both jit_entry and jit_launch_kernel. Maybe over here rename variable to `JitLaunchKernelFnTy`?

pass/ProteusPass.cpp
ggeorgakoudis: You suggestion is good but I want to address later encapsulating interface function creation. Can you open an issue?

pass/ProteusPass.cpp
ggeorgakoudis: Yes, makes the code clearer. As per your comment above, it makes sense to build encapsulating methods that are smartly named

scripts/gitlab/ci-build-test.sh
```diff
@@ -23,16 +23,16 @@ if [ "${CI_MACHINE}" == "lassen" ]; then
   LLVM_INSTALL_DIR=$(llvm-config --prefix)
   CMAKE_OPTIONS_MACHINE=" -DENABLE_CUDA=on -DCMAKE_CUDA_ARCHITECTURES=70 -DCMAKE_CUDA_COMPILER=$LLVM_INSTALL_DIR/bin/clang++"
 elif [ "${CI_MACHINE}" == "tioga" ]; then
-  ml load rocm/5.7.1
+  ml load rocm/${PROTEUS_CI_ROCM_VERSION}
 
-  LLVM_INSTALL_DIR=/opt/rocm-5.7.1/llvm
+  LLVM_INSTALL_DIR=${ROCM_PATH}/llvm
```
koparasy, line 28: Check if ${ROCM_PATH} is set and print the actual value. Can be beneficial to debug issues at some point.

scripts/setup-amd.sh
```diff
@@ -0,0 +1,24 @@
...
+    ROCM_VERSION="5.7.1"
+else
+    ROCM_VERSION=$1
+fi
+
+ml load rocm/${ROCM_VERSION}
+
+LLVM_INSTALL_DIR=${ROCM_PATH}/llvm
+
+mkdir build-rocm-${ROCM_VERSION}
```
koparasy, line 13: I don't like the file creation path. Add a $(hostname) field or ${SYS_TYPE} if this is only for internal systems. When we add additional clusters this can be an issue.

scripts/setup-amd.sh
```diff
@@ -0,0 +1,24 @@
...
+pushd build-rocm-${ROCM_VERSION}
+
+cmake .. \
+-DLLVM_INSTALL_DIR=${LLVM_INSTALL_DIR} \
+-DCMAKE_C_COMPILER=${LLVM_INSTALL_DIR}/bin/clang \
+-DCMAKE_CXX_COMPILER=${LLVM_INSTALL_DIR}/bin/clang++ \
+-DENABLE_HIP=on \
+-DCMAKE_INSTALL_PREFIX=../install-rocm-${ROCM_VERSION} \
```
koparasy, line 21: Same as above.

koparasy (changes requested):
Some are minor. Overall I think the pass would benefit by heaving some clear methods in `ProteusJitPassImpl` that show the interface to the runtime system. It will make it easier to read.

scripts/setup-amd.sh
ggeorgakoudis: I'll defer that, it's up to the user. We don't use the script for CI automation.

scripts/setup-amd.sh
ggeorgakoudis: Same as above

ggeorgakoudis:
**Note**: The CI tests on tioga will be failing for 6.2.1 and this is expected

tbennun (approved):
LGTM

koparasy (approved):

Issue #47, opened 2024-12-05
Fix the ABI of `__jit_launch_kernel` which depends on the host platform


ggeorgakoudis:
The problem is that the ABI for `[hip|cuda]LaunchKernel` (and correspondingly for `__jit_launch_kernel`) depends on the host platform.

For our Power9 host + NVIDIA GPU, cudaLaunchKernel takes grid size and block size arguments as structs of type [2 x i64] whereas for an x86_64 host + NVIDIA GPU (perlmutter) the structs are scalarized to two arguments of i64 plus i32 type.

We should detect the host ABI and modify ProteusPass and CompilerInterface modules.

PR #48, opened 2024-12-05
Support rdc device funcs

### LINKED ISSUE #40: Jitting kernels that call device functions defined in different translation unit breaks proteus
When JIT GPU kernels which call a device function implemented in a different translation unit proteus fails.  PR: #39 tests for this case and currently is marked as `WILL_FAIL`.

### LINKED ISSUE #45: Refactor pass implementation
The current pass implementation is hard to parse. We should implement some class at least that reflects the expectation of the runtime library.

### LINKED ISSUE #46: Tests on tioga for rocm 6.2.1/LLVM 18 fail
Test failure on rocm 6.2.1 on tioga:
```
The following tests FAILED:
	 21 - kernel_host_device_jit.HIP (Failed)
	 38 - kernel_host_device_jit.HIP.rdc (Failed)
```

The `kernel_host_device_jit` test creates a host JIT function that calls a kernel. The issues are:
1. In LLVM 18, ORC JIT does not make globals on custom sections externally visible
2. In LLVM 18, HIP creates a mangled global (`__hip_fatbin_<hash>`) to store the hip fatbin when Proteus using ORC JIT expects the global to be named simply `__hip_fatbin`.

We expect to solve that moving fatbin extraction through instrumentation. Relates to #39, #40

ggeorgakoudis:
Closes #40 
Closes #45
Closes #46


lib/JitEngineDevice.hpp
```diff
@@ -78,21 +82,31 @@ template <typename ImplT> class JitEngineDevice : protected JitEngine {
...
+    SmallVector<std::string> LinkedModuleIds;
+  };
 
-  void insertRegisterFunction(const void *HostAddr, char *FunctionName,
-                              int32_t *RCIndices, int32_t NumRCs) {
-    JITKernelInfoMap[HostAddr] = JITKernelInfo(FunctionName, RCIndices, NumRCs);
-  }
+  void *CurHandle = nullptr;
+  DenseMap<void *, BinaryInfo> HandleToBinaryInfo;
+  DenseMap<void *, void *> KernelToHandleMap;
```
koparasy, line 108: Maybe better name.

lib/JitEngineDeviceHIP.cpp
```diff
@@ -162,6 +208,16 @@ JitEngineDeviceHIP::codegenObject(Module &M, StringRef DeviceArch) {
...
+#if 0
+  for (auto &KV : LinkedModules) {
+    auto DeviceBitcode = KV.second;
+    hiprtcErrCheck(
+        hiprtcLinkAddData(hip_link_state_ptr, HIPRTC_JIT_INPUT_LLVM_BITCODE,
+                          (void *)DeviceBitcode.data(), DeviceBitcode.size(),
+                          "", 0, nullptr, nullptr));
+  }
+#endif
+
```
koparasy, line 220: Delete.

pass/ProteusPass.cpp
```diff
@@ -133,27 +132,32 @@ class ProteusJitPassImpl {
     // Host compilation
     // ================
 
+    instrumentRegisterLinkedBinary(M);
+    instrumentRegisterFatBinary(M);
+    instrumentRegisterFatBinaryEnd(M);
+    instrumentRegisterVar(M);
```
koparasy, line 138: We don't have a test for this yet.

pass/ProteusPass.cpp
```diff
@@ -133,27 +132,32 @@ class ProteusJitPassImpl {
...
         continue;
-      }
 
       emitJitModuleHost(M, JFI);
       emitJitEntryCall(M, JFI);
     }
 
+    DEBUG(dbgs() << "=== Post Original Host Module\n"
+                 << M << "=== End Post Original Host Module\n");
+
```
koparasy, line 160: Remove.

pass/ProteusPass.cpp
```diff
@@ -738,37 +637,20 @@ class ProteusJitPassImpl {
...
-    GlobalVariable *FatbinWrapper = nullptr;
-    FatbinWrapper = M.getGlobalVariable(FatbinName, true);
-    if (!FatbinWrapper)
-      FATAL_ERROR(
-          "Expected hip fatbinary wrapper global, check "
-          "ENABLE_CUDA|ENABLE_HIP compilation flags for ProteusJitPass");
-
-    FunctionType *JitEntryFnTy = nullptr;
+  FunctionCallee getJitLaunchKernelFn(Module &M) {
+    FunctionType *JitLaunchKernelFnTy = nullptr;
```
koparasy, line 641: Since this is a refactor once more. I would add a TODO/FIXME note here and mention the ABI issue

pass/ProteusPass.cpp
koparasy, line 138: This needs to be guarded for code that contains device stubs.

koparasy: I checked also the function `hasDeviceLaunchKernelCalls` and is not sufficient. You need to instrument when __hip/cuda_module_ctor is being called.

pass/ProteusPass.cpp
```diff
@@ -868,6 +742,135 @@ class ProteusJitPassImpl {
...
+        "__jit_register_fatbinary", JitRegisterFatbinaryFnTy);
+
+    return JitRegisterFatbinaryFn;
+  }
+
+  void instrumentRegisterFatBinary(Module &M) {
+    Function *F = nullptr;
+
+    if (!RegisterFatBinaryName)
+      return;
```
koparasy, line 759: This is a convoluted logic. Isn't there a better way to skip this code? Or, better never call this function if there is nothing todo.

pass/ProteusPass.cpp
```diff
@@ -868,6 +742,135 @@ class ProteusJitPassImpl {
...
+
+    F = M.getFunction(RegisterFatBinaryName);
+    if (!F)
+      return;
+
+    FunctionCallee JitRegisterFatBinaryFn = getJitRegisterFatBinaryFn(M);
```
koparasy, line 765: In the IRs that I am looking at this call is never being made when enabling rdc. It is only generated by the link.stub. I think you need access to the individual models (fatbins) as well.

pass/ProteusPass.cpp
```diff
@@ -868,6 +742,135 @@ class ProteusJitPassImpl {
...
+    // link.stub. It's not strictly necessary since this module will not have a
+    // device bitcode to pull and we skip at runtime.
+    if (!M.getGlobalVariable("__cuda_fatbin_wrapper", /*AllowInternal=*/true)) {
+      DEBUG(dbgs() << "Skip " << M.getSourceFileName() << "\n";)
+      return;
+    }
+
+    FunctionCallee JitRegisterLinkedBinaryFn = getJitRegisterLinkedBinaryFn(M);
+
+    for (auto &F : M.getFunctionList()) {
```
koparasy, line 843: Instead of iterating all functions, get the handle `__cuda_module_ctor`. Inside that function __cudaRegisterLinkedBinary is being called. The constructor is small, so in principle it should be faster.

pass/ProteusPass.cpp
```diff
@@ -868,6 +742,135 @@ class ProteusJitPassImpl {
...
+        if (!CB)
+          continue;
+
+        IRBuilder<> Builder(CB);
+        std::string GVName = getJitBitcodeUniqueName(M);
+        DEBUG(
+            dbgs() << "Instrument register linked binary to extract bitcode GV "
+                   << GVName << "\n");
+        auto *Arg = Builder.CreateGlobalString(GVName);
+        Builder.CreateCall(JitRegisterLinkedBinaryFn, {Arg});
```
koparasy, line 858: This will be deleted by entering the ctor body.

tests/gpu/device_func.cpp
```diff
@@ -0,0 +1,6 @@
+#include "gpu_common.h"
+#include <cstdio>
+
+__device__ void device_function(int a) { printf("device_func %d\n", a); }
+
+__global__ void kernel2() { printf("Kernel2\n"); }
```
koparasy, line 6: Either delete `kernel2` or create another test that has only device functions. 

The inclusion of a kernel, will enable your pass to perform additional actions (e.g. there is kernel stub and it will instrument it). But we need to test the case that there is no such stub.

koparasy: Can you also change:

`__device__ void device_function(int a) { printf("device_func %d\n", a); } ` 

to 

__device__ void device_function(int *a) {*a  = 123;}

Thus avoiding the TU to contain an external function?

koparasy (changes requested):
This overall looks good. The runtime portion is pretty clean. The pass requires work clean up.

1. Fix the comment on the tests.
2. I suggest we take advantage of the fact that clang will always insert a call `__vendor_module_ctor` that will be called by the host if the device side contains anything interesting for us. I would use this to clean up the additions. (e.g. you can detect whether you need to perform host-register related actions based on the existence of the function).

You don't need to fix in this PR the issue with external functions. But fixing 2 should (in principle) pave the way to that fix.

Thank you

lib/JitEngineDevice.hpp
ggeorgakoudis: Defer for later refactoring

pass/ProteusPass.cpp
ggeorgakoudis: There is an issue #47

pass/ProteusPass.cpp
ggeorgakoudis: that's our alternative for for #ifdefs. Defer, refactoring

tests/gpu/device_func.cpp
ggeorgakoudis: Why avoid the external function? I'm using it to check using lit

pass/ProteusPass.cpp
ggeorgakoudis: Proteus compiles the link.stub. I'll consider it resolved.

pass/ProteusPass.cpp
ggeorgakoudis: This is a good suggestion. I'll defer for iteration 2 and open an issue.

koparasy (approved):

PR #57, opened 2024-12-10
Use position-independent code

tbennun:
This makes the library compatible with other projects.

Are there any downsides?


CMakeLists.txt
```diff
@@ -7,6 +7,8 @@ project(Proteus
 set(CMAKE_CXX_STANDARD 17 CACHE STRING "")
 set(CMAKE_CXX_STANDARD_REQUIRED YES)
 set(CMAKE_CXX_EXTENSIONS NO)
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
ggeorgakoudis, line 10: Create an `ENABLE_PIC` cmake option to optionally set it

ggeorgakoudis:
@tbennun you're building a shared library and I think that's only when it must be PIC. Let's make it optional under a cmake option. There is a performance overhead because of the indirection, which we can avoid unless it's needed.

CMakeLists.txt
tbennun: Done.

ggeorgakoudis (approved):

davidbeckingsale:
Why not make this conditional on `BUILD_SHARED_LIBS`? A user could also manually set `CMAKE_POISITION_INDEPEDENT_CODE` on the command line, and we should respect that by not overwriting with the `set` command

davidbeckingsale (changes requested):
See other comment.

ggeorgakoudis:
> Why not make this conditional on `BUILD_SHARED_LIBS`? A user could also manually set `CMAKE_POISITION_INDEPEDENT_CODE` on the command line, and we should respect that by not overwriting with the `set` command

Thanks for double-checking @davidbeckingsale. The use-case is building libproteus as a static library but linking it with a shared library (@tbennun please correct me if I'm wrong). So, IIUC we don't need an extra option to support this use-case, just that the user needs to compile proteus with `CMAKE_POSITION_INDEPENDENT_CODE` set? That sounds reasonable as we deploy by source now and we could avoid the extra option (`ENABLE_PIC`). @tbennun thoughts? Does that work for you?

tbennun:
I’ll close this for now.

PR #63, opened 2024-12-12
Fix performance regression for device compileAndRun

### LINKED ISSUE #59: Performance regression for overhead in `compileAndRun`
For convenience we extract the types of runtime arguments from the bitcode. This means every time we need to parse the IR and extract them. We should avoid this overhead

ggeorgakoudis:
- Extract runtime constant types in the pass
- Avoid bitcode parsing unless compiling
- Link binaries for CUDA RDC in object codegen
- Add kernel repeat test

Closes #59


lib/CompilerInterfaceTypes.h
```diff
@@ -14,6 +14,19 @@
 #include <cstring>
 #include <stdint.h>
 
+namespace proteus {
+
+enum RuntimeConstantTypes : int32_t {
+  BOOL = 1,
```
johnbowen42, line 20: This is a great addition, I should've done this in the first place

lib/CompilerInterfaceTypes.h
```diff
@@ -14,6 +14,19 @@
 #include <cstring>
 #include <stdint.h>
 
+namespace proteus {
+
+enum RuntimeConstantTypes : int32_t {
```
johnbowen42, line 19: why inherit from, int32_t, just to enforce size?

lib/JitEngineDevice.hpp
```diff
@@ -51,20 +53,22 @@ using namespace llvm;
...
-      RCIndices.push_back(_RCIndices[I]);
-    }
-  }
-  JITKernelInfo() : Name(nullptr), NumRCs(0), RCIndices() {}
-  auto getName() const { return Name; }
-  auto getRCIndices() const { return RCIndices; }
-  auto getNumRCs() const { return NumRCs; }
+  JITKernelInfo(char const *Name, int32_t *RCIndices, int32_t *RCTypes,
+                int32_t NumRCs)
+      : Name(Name), RCIndices{ArrayRef{RCIndices, static_cast<size_t>(NumRCs)}},
```
johnbowen42, line 63: nice use of the constructors here

johnbowen42 (approved):
LGTM.  This improves not only performance but also readability and simplifies the code IMO.

lib/CompilerInterfaceTypes.h
ggeorgakoudis: Yes, to match the expected data type from the pass

lib/CompilerInterfaceTypes.h
ggeorgakoudis: Yes, to match the expected data type from the pass

tbennun (approved):
LGTM, solves my performance issue as well.

PR #64, opened 2024-12-12
Improve setting launching bounds on preset kernels

### LINKED ISSUE #27: Test setting launch bounds on a kernel with preset ones
We need a test to verify our approach for setting launch bounds works on a kernel with preset launch bounds. For CUDA, it should be possible to get and modify an existing metadata operand encoding launch bounds. Similarly, for HIP, using function attributes.

_Originally posted by @ggeorgakoudis in https://github.com/ggeorgakoudis/proteus/pull/15#discussion_r1823334170_

ggeorgakoudis:
- Replace metadata if already set (CUDA)
- Add test kernel_preset_bounds

Closes #27


lib/JitEngineDeviceCUDA.cpp
```diff
@@ -120,10 +121,26 @@ void JitEngineDeviceCUDA::setLaunchBoundsForKernel(Module &M, Function &F,
...
+    // Expecting 3 operands ptr, desc, i32 value.
+    assert(MetadataNode->getNumOperands() == 3);
+
+    auto *PtrMetadata = MetadataNode->getOperand(0).get();
+    auto *DescMetadata = MetadataNode->getOperand(1).get();
+    if (PtrMetadata == FuncMetadata && MaxntidxMetadata == DescMetadata) {
+      MetadataNode->replaceOperandWith(2, MaxThreadsMetadata);
+      return;
+    }
+  }
```
koparasy, line 140: Is this what we want? Are we going to override the developers intend? This sounds as bad practice. @davidbeckingsale what do you think?

ggeorgakoudis: Yes, we are overriding with a value that is correct to achieve greater optimization

tests/gpu/kernel_preset_bounds.cpp
```diff
@@ -0,0 +1,26 @@
...
+// Second run uses the object cache.
+// RUN: ./kernel_preset_bounds.%ext | FileCheck %s --check-prefixes=CHECK,CHECK-SECOND
+// clang-format on
+#include <climits>
+#include <cstdio>
+
+#include "gpu_common.h"
+
+__global__ __attribute__((annotate("jit")))
+__launch_bounds__(128, 4) void kernel() {
```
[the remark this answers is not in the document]
koparasy: Further, I would consider having the bounds as <1, 1>. To match the call site and indirectly test case in which proteus does not call the function correctly (for whatever reason).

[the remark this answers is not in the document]
ggeorgakoudis: No, it's not mandatory. The `assert` will not fail because this is the format of the metadata.

koparasy (changes requested):
Nice work. Nits

lib/JitEngineDeviceCUDA.cpp
koparasy: You don't know that though. What if the user has auto-tuned the code and knows that these values optimize the code? There is no guarantee on what you do there.

ggeorgakoudis: It's possible to turn off this optimization with an env var. Down the road we can add annotation extensions to turn it on/off per kernel

tests/gpu/kernel_preset_bounds.cpp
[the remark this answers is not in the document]
ggeorgakoudis: I'm using different launch bounds than the actual launch dimensions because I can visually check the IR if the optimization happened. Checking will be automated in the future.

lib/JitEngineDeviceCUDA.cpp
koparasy: IMO this is not intuitive. The developer says:
```
__global__ __attribute__((annotate("jit")))
__launch_bounds__(128, 4)
```
this directly translates to  explicit control to turn it on/off per kernel with no additional support from us.

ggeorgakoudis: User can be mistaken, Proteus isn't 😝

ggeorgakoudis:
Rebased ☝️

ggeorgakoudis:
@koparasy @davidbeckingsale Given the discussion in slack on the necessity to override launch bounds for HIP, can we approve this PR?

koparasy (approved):

PR #66, opened 2024-12-13
Add Logger for file output logs

### LINKED ISSUE #24: Implement a unified logging interface
For logging, we use LLVM's `dbgs()` extensively for debug output while the rest of logging is a mix of std, printf, and possibly LLVM's `outs()`. There were issues with output re-ordering depending on the different streams flushing. We need a better logger.

_Originally posted by @ggeorgakoudis in https://github.com/ggeorgakoudis/proteus/pull/15#discussion_r1819959737_

ggeorgakoudis:
Store debug output in files

Closes #24

### DIFF
--- a/common/Logger.hpp
+++ b/common/Logger.hpp
@@ -1,0 +1,32 @@
+#include <filesystem>
+#include <llvm/Support/Error.h>
+#include <llvm/Support/FileSystem.h>
+#include <llvm/Support/raw_ostream.h>
+#include <stdexcept>
+#include <system_error>
+
+namespace proteus {
+
+class Logger {
+public:
+  static llvm::raw_ostream &logs(const std::string &Name) {
+    static Logger SingletonLogger{Name};
+    return SingletonLogger.OutStream;
+  }
+
+private:
+  const std::string LogDir = ".proteus-logs";
+  bool DirExists;
+  std::error_code EC;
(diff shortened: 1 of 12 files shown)


ggeorgakoudis:
Rebase ☝️

ggeorgakoudis:
Rebase ☝️

davidbeckingsale:
I think you should add PID to the log filename to help differentiate when running multiprocess, or just running same binary over and over.

davidbeckingsale:
Also maybe we could/should consider a third-party dependency for this: https://github.com/gabime/spdlog

ggeorgakoudis:
> I think you should add PID to the log filename to help differentiate when running multiprocess, or just running same binary over and over.

Does it work to open an issue for this? I intended this to be a very basic commit that just introduces the functionality to move from monitor printouts to a file, even if overwritten. This still serves the use-case of manual. visual inspecting where we run one compilation/execution and want to see the output.

ggeorgakoudis:
> Also maybe we could/should consider a third-party dependency for this: https://github.com/gabime/spdlog

@davidbeckingsale I wasn't aware of `spdlog` and it looks awesome! However, I'd like to keep dependencies to a minimum until we have a robust, well-integrated core. Can we again open an issue for forward development?

davidbeckingsale (approved):

PR #72, opened 2024-12-18
Computes a SHA256 hash per device compilation unit and uses for hashing

koparasy:


lib/JitEngineDevice.hpp
```diff
@@ -592,31 +641,63 @@ void JitEngineDevice<ImplT>::registerLinkedBinary(
...
+
+                Function *F = Mod.getFunction(Symbol.getKey());
+                if (!F)
+                  continue;
+
+                // Internalize functions, the JIT module is
+                // self-contained.
+                F->setLinkage(GlobalValue::InternalLinkage);
+              }
+            }))
```
koparasy, line 700: @ggeorgakoudis I have encountered a weird case on the `linkInModule` functionality. Whether the Linking process will succeed depends on the order of the modules in the `LinkedModules` vector. Is this expected?

koparasy: To verify, I went on the main branch and modified the code:

```
for (auto &LinkedM : LinkedModules) 
```

to 

```
for (auto &LinkedM : llvm::reverse(LinkedModules))
```

and I am getting:
```
ERROR @ /dev/shm/proteus/lib/JitEngineDeviceCUDA.cpp:220 ->  3
Aborted
```

From my investigation, I think the order of Modules in LinkedModules should respect the dependency traversal. In other words, we should not add in the Linker functions before their users are added. Is this something you are aware off? Or did the tests just pass because of random order of registration?

koparasy:
@ggeorgakoudis Don't wast time in looking in the rest of the PR, it is a draft and I will work it out later. Let me know about the comment though.

lib/JitEngineDevice.hpp
ggeorgakoudis: Good observation. The IRLinker is stateful and the order matters since we do "link as needed" and internalize symbols through the chain of linking. The current linking is correct because the order in LinkerModuleIds should always have the Module of the kernel-to-JIT first, followed by other device Modules to link in. The error that you see is because the NVPTX compiler failed, most probably due to a missing definition in the final linked Module. I agree this is brittle and we could easily make it order-independent if we just link in everything without internalization, then perform internalization inside the specializeIR stage. Can you open an issue? I'll fix this in a jiff.

koparasy: I think internalizing later is better, there are going to be cases of some cyclic weird dependency and the order of the modules will be hard to describe unless we move the internalization later.

lib/JitEngineDeviceCUDA.hpp
```diff
@@ -79,13 +80,18 @@ class JitEngineDeviceCUDA : public JitEngineDevice<JitEngineDeviceCUDA> {
     return "llvm.nvvm.read.ptx.sreg.tid.z";
   };
 
+  static bool HashSection(StringRef sectionName) {
+    static DenseSet<StringRef> Sections{".nv_fatbin", ".nvFatBinSegment",
+                                        "__nv_relfatbin", "__nv_module_id"};
```
koparasy, line 85: We only need the `nv_fatbin`

PR #78, opened 2024-12-19
Fix multiple annotations with RDC

davidbeckingsale:
- Add test `multi_file_launcher`
- Fix by removing overconservative fatal errors

### DIFF
--- a/lib/JitEngineDevice.hpp
+++ b/lib/JitEngineDevice.hpp
@@ -585,6 +585,8 @@ void JitEngineDevice<ImplT>::registerFunction(void *Handle, void *Kernel,
       << "Register function " << Kernel << " To Handle " << Handle << "\n");
   assert(!KernelToHandleMap.contains(Kernel) &&
          "Expected kernel inserted only once in the map");
+  //if (KernelToHandleMap.contains(Kernel))
+  //  return;
   KernelToHandleMap[Kernel] = Handle;
 
   JITKernelInfoMap[Kernel] =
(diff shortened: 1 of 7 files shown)


ggeorgakoudis (approved):
LGTM

koparasy:
This is kind of weird. I would expect the current implementation of the cache to find the same kernel. As it doesn't account for the translation unit location. Aren't the kernel names of these codes the same?

ggeorgakoudis:
> This is kind of weird. I would expect the current implementation of the cache to find the same kernel. As it doesn't account for the translation unit location. Aren't the kernel names of these codes the same?

The module unique ids by LLVM are different

koparasy:
> > This is kind of weird. I would expect the current implementation of the cache to find the same kernel. As it doesn't account for the translation unit location. Aren't the kernel names of these codes the same?
> 
> The module unique ids by LLVM are different

I don't understand that either. Do these 2 device codes have different external dependencies? I thought the unique id captures the external dependencies of the current translation unit.

ggeorgakoudis:
> > > This is kind of weird. I would expect the current implementation of the cache to find the same kernel. As it doesn't account for the translation unit location. Aren't the kernel names of these codes the same?
> > 
> > 
> > The module unique ids by LLVM are different
> 
> I don't understand that either. Do these 2 device codes have different external dependencies? I thought the unique id captures the external dependencies of the current translation unit.

Check the test, one translation unit will export `main` the other `foo`, results in different module ids

koparasy:
Both foo, and main are cpu codes. Not device. So, do we expect to capture both host and device externals?

ggeorgakoudis:
> Both foo, and main are cpu codes. Not device. So, do we expect to capture both host and device externals?

The rationale is to create a module unique id for the translation unit that includes the kernel. Host code will also have the kernel stub. The module unique id is not ideal anyway and will be removed.

Issue #79, opened 2024-12-19
Pass registration order


koparasy:
Why do we register the pass early? If we register it late the embedded IR will include any optimizations taking place during middle end e.g. fast-math, vectorize-slp, vectorize-loops etc ...

I recall there was a linking issue, do we have a test for this issue?

PR #81, opened 2024-12-19
Register proteus pass late

koparasy:
Registers pass late as discussed in #79 . 

Pending tasks:
- [x] Verify correctness in systems and architectures
- [ ] Verify constant threadDim/BlockDim specialization ( PR #42 )

### DIFF
--- a/pass/ProteusPass.cpp
+++ b/pass/ProteusPass.cpp
@@ -1154,11 +1154,11 @@ llvm::PassPluginLibraryInfo getProteusJitPassPluginInfo() {
 
     // PB.registerPipelineStartEPCallback(
     // PB.registerOptimizerLastEPCallback(
-    PB.registerPipelineEarlySimplificationEPCallback(
-        [&](ModulePassManager &MPM, auto) {
-          MPM.addPass(ProteusJitPass{false});
-          return true;
-        });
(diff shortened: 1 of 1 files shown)


davidbeckingsale:
Is there concern this will break any of our analysis/transformations? e.g. lambda specialization stuff looks for very specific patterns

koparasy:
> Is there concern this will break any of our analysis/transformations? e.g. lambda specialization stuff looks for very specific patterns

It may, I have not checked the patterns we are currently "catching". Do we have tests for these patterns?

davidbeckingsale:
Test programs yes, but no IR checking

koparasy:
this is a low priority PR, and I am still testing things out. Deciding on which stage we are going to run the pipeline is important and we should do it asap. In terms of performance, it makes more sense to register it late.

koparasy:
I will take that back. It is possible to run the annotation analysis early in the pipeline, and the registering of the LLVM IR files late in the pipeline. This may give us both capabilities.

davidbeckingsale:
If the GitLab passes then I'm not worried :)

koparasy:
Passed :)

ggeorgakoudis:
An issue with late registration is that we may miss JIT annotated functions that have been inlined

koparasy:
That's a good point

PR #87, opened 2024-12-23
Allow variadic args in hash function

davidbeckingsale:


ggeorgakoudis (approved):
LGTM

ggeorgakoudis:
Also hashes GridDim, BlockDim as we should

PR #92, opened 2025-01-07
Reorder blockId/threadId

koparasy:
@ggeorgakoudis I would like to write a test, my current approach would be to have a kernel that prints out BlockDims/GridDims and we grep them.

This will not verify that the optimization is correct, but it should at least verify that we are not breaking any program assumptions.

Do you have a better idea?


ggeorgakoudis:
> @ggeorgakoudis I would like to write a test, my current approach would be to have a kernel that prints out BlockDims/GridDims and we grep them.
> 
> This will not verify that the optimization is correct, but it should at least verify that we are not breaking any program assumptions.
> 
> Do you have a better idea?

Sounds good, I can't think of something better.

ggeorgakoudis (approved):
LGTM

PR #99, opened 2025-01-08
Add cpp linter workflow

ggeorgakoudis:
An improved way to add clang-format and clang-tidy checks by showing required changes as diff comments and in the convo.

The `.clang-tidy` configuration matches LLVM's.


ggeorgakoudis:
For a demo, this PR has useful visuals: #94. We will have the `Cpp-linter Review` with suggested changes in the discussion and also in the diff.

davidbeckingsale:
Is there a way to auto apply changes, perhaps with a / comment on the PR?

ggeorgakoudis:
> Is there a way to auto apply changes, perhaps with a / comment on the PR?

I think we can do that for `clang-format` changes, but for `clang-tidy` it's more nuanced. The `clang-tidy` review will suggest changes that may need manual editing (e.g., name of variables). The nice thing about `cpp-linter` is that it adds a "commit" button on suggested changes so it's easy to apply thing through github's GUI

ggeorgakoudis:
@davidbeckingsale Is this good to go?

davidbeckingsale (approved):

PR #101, opened 2025-01-08
Extract device bitcode once optimization

koparasy:
Avoid re-parsing the bitcode in a memory buffer. Store linked IR modules in the kernel info map.


ggeorgakoudis (approved):
LGTM. I believe some refactoring is possible to move extraction earlier and avoid duplication, will open an issue.

PR #108, opened 2025-01-13
Add faster codegen for HIP without using RTC API

### LINKED ISSUE #60: Running O3 twice on HIP codes regression
We run O3 twice once through `optimizeIR` and once through HIP RTC interfaces in HIP code. It may impact code generation negatively or positively. Need to investigate.

ggeorgakoudis:
HIP RTC codegen is relatively slow because it invokes the clang-driver. This PR adds an alternative codegen that creates the device binary programmatically.

- Add alternative codegen for HIP devices
- Add env var to enable the alternative codegen by setting ENV_PROTEUS_USE_HIP_RTC_CODEGEN to `off`
- Add scale-out tests with 100 files (1 kernel and 1 device function per kernel) to compile and link using RDC, with/without a device global variable

> [!NOTE]
> To reviewers: ignore scale100* directories for the test, they are auto-generated

Closes #60


lib/JitEngine.cpp
```diff
@@ -118,6 +118,8 @@ JitEngine::JitEngine() {
       getEnvOrDefaultBool("ENV_PROTEUS_SPECIALIZE_ARGS", true);
   Config.ENV_PROTEUS_SPECIALIZE_DIMS =
       getEnvOrDefaultBool("ENV_PROTEUS_SPECIALIZE_DIMS", true);
+  Config.ENV_PROTEUS_USE_HIP_RTC_CODEGEN =
+      getEnvOrDefaultBool("ENV_PROTEUS_USE_HIP_RTC_CODEGEN", true);
```
davidbeckingsale, line 122: Do we want to default to false?

ggeorgakoudis: > Do we want to default to false?

I'm struggling to make a decision. HIP RTC might generate faster device code assuming the RTC optimization pipeline is better than our own, so I left on by default.

ggeorgakoudis: Made my mind. Let's keep it as is, so default HIP RTC, until we're confident our own codegen is good. Configuration is an env var so easy to switch it a runtime without re-compilation.

ggeorgakoudis:
@davidbeckingsale We discussed offline (with Dinos too) whether to include the scale100 test. I'd like to have them in because it makes easy to run and debug scaling issues with them. I know they're a lot of files but the byte suze in git history is very small since they are text. We can remove them later if we re-consider.

Is there anything pending? Can we merge?

davidbeckingsale (approved):

davidbeckingsale:
> @davidbeckingsale We discussed offline (with Dinos too) whether to include the scale100 test. I'd like to have them in because it makes easy to run and debug scaling issues with them. I know they're a lot of files but the byte size in git history is very small since they are text. We can remove them later if we re-consider.
> 
> Is there anything pending? Can we merge?

All good, I agree keeping the tests is useful

PR #112, opened 2025-01-15
Detects AMD Thread/Block/Grid Dim intrinsics and specializes with them

### LINKED ISSUE #111: Specialize dims for HIP RDC may not apply
We expect that compiling HIP with RDC creates an LTO module of AOT-optimized TU bitcodes. Calls to dim getters should have been lowered to intrinsics instead of C++ symbols. Specialize dims for HIP detects C++ symbols.

koparasy:
Fixes  #111


ggeorgakoudis (approved):
LGTM

Issue #96, opened 2025-01-07
bug: RAJA reduction segfault while compiling for multiple HIP targets


johnbowen42:
When RAJAPerf is compiled with the `-DCMAKE_HIP_ARCHITECTURES=gfx90a,gfx906`, RAJA `forall` loops with reductions segfault.  This could be an error related to proteus specializing for the incorrect architecture

ggeorgakoudis:
@johnbowen42 Is this still an issue? Can I close it?

johnbowen42:
this is still an issue, but lower priority

ggeorgakoudis:
Is it failing when we compile for a single AMD target device?

davidbeckingsale:
It works for a single arch

ggeorgakoudis:
I suspect this issue here is a duplicate of #115

ggeorgakoudis:
@johnbowen42 Can you check that PR #116 fixes this issue and close it?

PR #116, opened 2025-01-16
Bugfix/add multi arch support

### LINKED ISSUE #115: Unable to handle multi-arch fatbinaries.
Applications that build for multi-gpu architectures fail to properly detect the fatbin for the running architecture. On mneme I detect at [runtime the architecture](https://github.com/Olympus-HPC/Mneme/blob/features/register-ir/src/MnemeRecordHIP.cpp#L182) and [skip](https://github.com/Olympus-HPC/Mneme/blob/features/register-ir/src/MnemeRecordHIP.cpp#L60) the ones that do not match the current architecture.

koparasy:
Closes #115 and #96. 

I went over the LLVM IR being loaded in cuda, and it always pulls the valid sm architecture.


scripts/gitlab/ci-build-test.sh
```diff
@@ -46,6 +46,7 @@ CMAKE_OPTIONS="-DLLVM_INSTALL_DIR=$LLVM_INSTALL_DIR"
 CMAKE_OPTIONS+=" -DCMAKE_C_COMPILER=$LLVM_INSTALL_DIR/bin/clang -DCMAKE_CXX_COMPILER=$LLVM_INSTALL_DIR/bin/clang++"
 CMAKE_OPTIONS+=${CMAKE_OPTIONS_MACHINE}
 CMAKE_OPTIONS+=" -DENABLE_DEBUG=${PROTEUS_CI_ENABLE_DEBUG} -DENABLE_TIME_TRACING=${PROTEUS_CI_ENABLE_TIME_TRACING}"
+CMAKE_OPTIONS+="-DCMAKE_HIP_ARCHITECTURES=\"gfx942;gfx90a\""
```
ggeorgakoudis, line 49: I'm not fond of testing by forcing the architecture (also on CUDA building?) on all tests. How does it work for CUDA?

Can we omit this test? It's an obvious fatal error.

ggeorgakoudis (changes requested):
Remove code duplication, not sure about testing.

scripts/gitlab/ci-build-test.sh
koparasy: In CUDA we specify the architecture explicitly. These are the 2 architectures we typically use proteus on and these are the respective architectures users use.

ggeorgakoudis: But we can specify multiple archs for CUDA too, no? We extract bitcode for CUDA through reading the device so I guess that's how we sidestep this issue.

Fair enough about HIP testing. Let's move the multi-arch option inside the tioga-only cmake options

ggeorgakoudis (approved):
LGTM

PR #129, opened 2025-01-22
Use upper case letters on setup scripts

koparasy:
#125 introduced bug on our setup scripts


ggeorgakoudis (commented):
What's the bug? Why is this needed?

koparasy:
Consider that this never happened

PR #132, opened 2025-01-23
bug: fix Cmake install typo

johnbowen42:


ggeorgakoudis:
Those lines got introduced here:
https://github.com/Olympus-HPC/proteus/commit/a30d426e55f438562a6d8a1d01d7729e66ad0fe9#diff-4803677bd280f4dded52c428e3c8ac6c618508dcabcbe880b68be85897c2ad41R115-R119

but I don't think we need them, since `proteusTargets.cmake` is already exported here:
https://github.com/Olympus-HPC/proteus/blob/main/CMakeLists.txt#L59-L60

We should just delete them?

koparasy:
We should delete them. Yes.

ggeorgakoudis (approved):

ggeorgakoudis:
Pending tests, merge

PR #137, opened 2025-01-28
Support device lambda capture specialization

davidbeckingsale:


lib/JitEngine.hpp
```diff
@@ -45,6 +46,9 @@ class JitEngine {
 
   bool isProteusDisabled() { return Config.ENV_PROTEUS_DISABLE; }
 
+  void pushJitVariable(RuntimeConstant RC);
```
ggeorgakoudis, line 49: reference?

lib/JitEngine.hpp
```diff
@@ -69,6 +73,33 @@ class JitEngine {
...
 
+inline
+SmallVector<RuntimeConstant, 8>& getPendingJitVariables() {
+  static SmallVector<RuntimeConstant, 8> PendingJitVariables;
+  return PendingJitVariables;
+}
+
+inline
+DenseMap<StringRef, SmallVector<RuntimeConstant, 4>>& getJitVariableMap() {
+  static DenseMap<StringRef, SmallVector<RuntimeConstant, 4>> JitVariableMap;
```
ggeorgakoudis, line 84: is it safe to use `StringRef` for the key without the ownership? I assume yes because it's a ref to the global we create in the ProteusPass? Add comment

lib/JitEngineDevice.hpp
```diff
@@ -512,6 +513,13 @@ void JitEngineDevice<ImplT>::specializeIR(Module &M, StringRef FnName,
         M, *F, RCIndices,
         ArrayRef<RuntimeConstant>{RC,
                                   static_cast<size_t>(NumRuntimeConstants)});
+  
+  if (! getJitVariableMap().empty() ) {
+    for (auto& F : M.getFunctionList()) {
+      if (llvm::demangle(F.getName()).find("lambda") != std::string::npos)
```
ggeorgakoudis, line 519: why don't we match the Symbol (key in the map) to call the transform only if needed?

lib/TransformLambdaSpecialization.hpp
```diff
@@ -55,12 +56,25 @@ static inline Constant *getConstant(LLVMContext &Ctx, Type *ArgType,
...
     PROTEUS_DBG(Logger::logs("proteus")
                 << "TransformLambdaSpecialization::transform" << "\n");
     PROTEUS_DBG(Logger::logs("proteus") << "\t args" << "\n");
-    for (auto &Arg : RC) {
+
+    const std::string Operator{llvm::demangle(F.getName())};
+    std::size_t Sep = Operator.find("::");
+    std::size_t SecondSep = Operator.find("::", Sep+1);
+
+    const std::string Symbol = Operator.substr(0, SecondSep);
```
ggeorgakoudis, line 69: optional: use `slice` on the `StringRef` of the function name to isolate the `Symbol`

lib/TransformLambdaSpecialization.hpp
```diff
@@ -69,7 +83,7 @@ class TransformLambdaSpecialization {
     for (User *User : LambdaClass->users()) {
       PROTEUS_DBG(Logger::logs("proteus") << *User << "\n");
       if (dyn_cast<LoadInst>(User)) {
```
ggeorgakoudis, line 85: optional: `isa<>` instead of `dyn_cast<>`

lib/TransformLambdaSpecialization.hpp
```diff
@@ -69,7 +83,7 @@ class TransformLambdaSpecialization {
...
-        for (auto &Arg : RC) {
+        for (auto &Arg : RC->second) {
```
ggeorgakoudis, line 86: Before loops assign `auto &RCVec = RC->second` for readability

pass/ProteusPass.cpp
```diff
@@ -1109,6 +1112,41 @@ class ProteusJitPassImpl {
...
+      DEBUG(Logger::logs("proteus-pass") << "registering lambda functions" << "\n");
+
+      SmallVector<Function *, 16> LambdaFunctions;
+      for (auto &F : M.getFunctionList()) {
+        // TODO: Demangle and search for the fully qualified proteus::jit_variable
+        // name.
+        if (F.getName().contains("register_lambda")) {
+          LambdaFunctions.push_back(&F);
+        }
+      }
```
ggeorgakoudis, line 1126: optional
```
if (!F.getName().contains...)
  continue;

LambdaFunctions.push_back(&F);
```
Typical LLVM-style and fewer braces

pass/ProteusPass.cpp
```diff
@@ -1109,6 +1112,41 @@ class ProteusJitPassImpl {
...
+
+      for (auto Function : LambdaFunctions) {
```
ggeorgakoudis, line 1128: `auto *Function`

surprised `clang-tidy` didn't pick this up, it should be the draft status

pass/ProteusPass.cpp
```diff
@@ -1109,6 +1112,41 @@ class ProteusJitPassImpl {
...
+        auto DemangledName = llvm::demangle(Function->getName());
+
+        std::size_t L = DemangledName.find("<");
+        std::size_t R = DemangledName.find(">"); 
+        const std::string Lambda = DemangledName.substr(L+1, R-L-2);
+        StringRef LambdaSymbol{Lambda};
```
ggeorgakoudis, line 1134: optional: `StringRef` with `slice`

lib/JitEngineHost.cpp
```diff
@@ -273,13 +273,12 @@ JitEngineHost::specializeIR(StringRef FnName, StringRef Suffix, StringRef IR,
         *M, *F, ArgPos,
         ArrayRef<RuntimeConstant>{RC,
                                   static_cast<size_t>(NumRuntimeConstants)});
-    if (!JitVariables.empty())
-      TransformLambdaSpecialization::transform(*M, *F, JitVariables);
+    if (!getJitVariableMap().empty())
```
ggeorgakoudis, line 276: Same here, do the matching and call the transform only if needed

lib/JitVariable.hpp
```diff
@@ -12,15 +12,21 @@
...
   std::memcpy(&RC, &v, sizeof(T));
   RC.Slot = pos;
   __jit_push_variable(RC);
 
   return v;
 }
+
+template <typename T>
+void register_lambda(T&& t, const char* Symbol="") {
+  __jit_register_fn(Symbol);
```
ggeorgakoudis, line 30: runtime check or assert that `Symbol` is not empty

pass/ProteusPass.cpp
```diff
@@ -250,10 +252,11 @@ class ProteusJitPassImpl {
       if (isDeviceCompilation(M)) {
         ModuleDeviceKernels = getDeviceKernels(M);
         if (!isDeviceKernel(Fn))
-          FATAL_ERROR(std::string{} + __FILE__ + ":" +
-                      std::to_string(__LINE__) +
-                      " => Expected the annotated Fn " + Fn->getName() +
-                      " to be a kernel function!");
+          return;
```
ggeorgakoudis, line 255: Why `return`? Shouldn't it be continue. There may other JIT kernels in the module.

ggeorgakoudis (commented):
LG, some stylistic comments and clarifications. Remove draft status and let's get this landed.

lib/JitEngineDevice.hpp
davidbeckingsale: Symbol is of the lambda class, not operator(), so we cannot do a direct match.

ggeorgakoudis:
Rebase ☝️

include/proteus/JitEngineDevice.hpp
```diff
@@ -489,6 +491,16 @@ void JitEngineDevice<ImplT>::specializeIR(Module &M, StringRef FnName,
         ArrayRef<RuntimeConstant>{RC,
                                   static_cast<size_t>(NumRuntimeConstants)});
 
+  if (!getJitVariableMap().empty()) {
+    for (auto &F : M.getFunctionList()) {
+      if (auto OptionalRCVec = matchJitVariableMap(F.getName())) {
+        TransformLambdaSpecialization::transform(M, F, OptionalRCVec.value());
+        getJitVariableMap().erase(F.getName());
```
davidbeckingsale, line 498: Better to keep iterator from the match and use that for erase, rather than use the key?

ggeorgakoudis: Avoid searching, right?

ggeorgakoudis:
Rebase and update ☝️

ggeorgakoudis (approved):

PR #139, opened 2025-01-30
Make the code path with disabled proteus faster

ggeorgakoudis:
Avoid the overhead looking into the KernelInfo map if proteus is disabled

### DIFF
--- a/lib/CompilerInterfaceDeviceInternal.hpp
+++ b/lib/CompilerInterfaceDeviceInternal.hpp
@@ -1,5 +1,7 @@
 #include "CompilerInterfaceDevice.h"
 
+#include "CoreDevice.hpp"
+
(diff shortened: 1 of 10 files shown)


koparasy (approved):

ggeorgakoudis:
Rebase ☝️

lib/CompilerInterfaceDeviceInternal.hpp
```diff
@@ -9,10 +11,19 @@ static inline auto __jit_launch_kernel_internal(void *Kernel, dim3 GridDim,
 
   using namespace llvm;
   using namespace proteus;
+
+  static bool IsProteusDisabled =
```
davidbeckingsale, line 15: Could be const?

davidbeckingsale (approved):

PR #145, opened 2025-02-08
Update HIP LTO handling to unblock specialization

ggeorgakoudis:
- Use LTO code for importing undefined globals at JIT linking, instead of replacing per TU module extracted bitcode. LTO includes code post-optimization, which blocks proteus specialization (e.g., due to inlining) if used as the JIT module.


ggeorgakoudis:
Rebase ☝️

davidbeckingsale (approved):

PR #149, opened 2025-02-11
Fix host specific ABI for __jit_launch_kernel

### LINKED ISSUE #47: Fix the ABI of `__jit_launch_kernel` which depends on the host platform

ggeorgakoudis:
Closes #47


koparasy:
This works, and it does what you want to do. Do we want to change the runtime? I would expect the proteus-runtime compiled to powerpc/x86 to have the correct ABI, without explicit handling. 

The pass needs to change, but again I don't think we should ifdef things. Isn't it enough to deduce the ABI by looking the original cuda/hipLaunchKernel?

ggeorgakoudis:
> This works, and it does what you want to do. Do we want to change the runtime? I would expect the proteus-runtime compiled to powerpc/x86 to have the correct ABI, without explicit handling.

You are right, that's why I left the "abundance of caution" comment.

> 
> The pass needs to change, but again I don't think we should ifdef things. Isn't it enough to deduce the ABI by looking the original cuda/hipLaunchKernel?

To find the ABI, I'll need to do dirty tricks with compiling and reading the LLVM IR from a helper source file (like we do for `struct RuntimeConstant`), which I loathe. Unless, you have a better idea.

PR #152, opened 2025-02-12
Fix type matching for lambdas with arguments 

tomstitt:
Loosen demangled name matching strictness to handle lambda launchers with arguments and print demangled name if check fails. Previously I'd get errors for something like the following for `... void kernel(int n, T LB)`:

`
fatal error: error in backend: .../proteus/src/pass/ProteusPass.cpp:266 => .../proteus/src/pass/ProteusPass.cpp:263 => Expected the annotated Fn _ZZ4mainENKUliE_clEi (main::'lambda'(int)::operator()(int) const) to be a kernel function or device lambda function!
`

I also added an example to the gpu lambda test.

### DIFF
--- a/tests/gpu/lambda.cpp
+++ b/tests/gpu/lambda.cpp
@@ -17,9 +17,24 @@ __global__ __attribute__((annotate("jit"))) void kernel(T LB) {
     LB();
 }
 
+template <typename T>
+__global__ __attribute__((annotate("jit"))) void kernel(int n, T LB) {
+  std::size_t i = blockDim.x * blockIdx.x + threadIdx.x;
+  if (i < n)
+    LB(i);
+}
+
 template <typename T> void register_run(T &&LB) {
   proteus::register_lambda(LB);
   kernel<<<1, 1>>>(LB);
+  gpuErrCheck(gpuDeviceSynchronize());
+}
+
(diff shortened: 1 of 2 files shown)


davidbeckingsale (approved):
Good catch! I wonder if we should use regex for string matching

tomstitt:
> Good catch! I wonder if we should use regex for string matching

Yes, I was wondering about that, it would also mean only one match, I'll try it !

ggeorgakoudis (commented):
LGTM.

What's failing in the test?

tests/gpu/lambda.cpp
```diff
@@ -46,6 +61,12 @@ int main(int argc, char **argv) {
 
   std::cout << "x[0] = " << x[0] << "\n";
 
+  register_run(
+      2, [ =, c = proteus::jit_variable(a) ] __device__
+      __attribute__((annotate("jit"))) (int i) { x[i] = c; });
+
+  std::cout << "x[0:1] = " << x[0] << "," << x[1] << "\n";
```
ggeorgakoudis, line 68: OK, I think it's this printout that's failing. We should change the `CHECK:` line

tomstitt: ah! yes, I did try locally, I will fix

PR #155, opened 2025-02-14
Fix segfault from unsafe DenseMap in global constructor

ggeorgakoudis:
- Use STL std::unordered_map instead of LLVM ADT DenseMap (which is unsafe during static initialization) to avoid memory corruption


davidbeckingsale (approved):
assuming this is because DenseMap is not safe to use during static initialization, would be worth adding that to the commit message.

PR #153, opened 2025-02-13
Refactor to provide a core external interface

ggeorgakoudis:
- Support for Mneme

Closes #23

PR #159, opened 2025-02-17
Create a Core Target in cmake

koparasy:
This exposes the header only proteus target to downstream users as a cmake `INTERFACE`. 

It resolves many issues that arise when using the previous proteus target. Some mentioned in #153 others relate to multiple LLVM symbol redefinitions.

Exposing proteusCore as a target does not bring the libproteus.so during linking which in turn brings in LLVM.


koparasy:
This was also tested with Mneme and alll Mneme tests pass

ggeorgakoudis:
That's a very cool idea and I think it fits well the Mneme use-case.

src/lib/CMakeLists.txt
```diff
@@ -130,6 +151,14 @@ install(
   LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
   ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}")
 
+install(
```
ggeorgakoudis, line 154: Why do we need this `install` for the proteusCore since it's header-only? It doesn't produce any binaries to install and the next `install` that copies headers is enough, no?

ggeorgakoudis:
Add also integration tests

src/lib/CMakeLists.txt
```diff
@@ -99,7 +115,9 @@ endif()
 
 if(PROTEUS_ENABLE_HIP)
   target_include_directories(proteus SYSTEM PRIVATE ${hip_INCLUDE_DIRS})
+  target_include_directories(proteusCore SYSTEM INTERFACE ${hip_INCLUDE_DIRS})
   target_compile_options(proteus PRIVATE -x hip)
+  target_compile_options(proteusCore INTERFACE -x hip)
```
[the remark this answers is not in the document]
koparasy: As it is now, anyone that uses HIP_SYMBOL(x) needs to be treated as a HIP code.

src/lib/CMakeLists.txt
koparasy: Here I am not installing the headers, I install the cmake target. I just pushed a cleaner version of the same functionality by installing multiple targets with the same command.

ggeorgakoudis: Yes but the `proteusCore` target does not create any RUNTIME, LIBRARY, or ARCHIVE artifacts, so what's the use?

PR #164, opened 2025-02-27
[Refactor] Move specializeIR to the core interface

ggeorgakoudis:
* I will open a todo issue on adding a configuration option for enabling/disabling lambda capture specialization


koparasy (approved):
LGTM

PR #167, opened 2025-02-28
Support asynchronous compilation

ggeorgakoudis:
Updates to tests will follow in a separate commit


include/proteus/CoreLLVM.hpp
```diff
@@ -134,14 +134,15 @@ inline void runOptimizationPassPipeline(Module &M, StringRef Arch,
...
-}
+struct InitLLVMTargets {
+  InitLLVMTargets() {
+    InitializeAllTargetInfos();
+    InitializeAllTargets();
+    InitializeAllTargetMCs();
+    InitializeAllAsmParsers();
+    InitializeAllAsmPrinters();
+  }
+};
```
koparasy, line 145: This seems as independent to the async, and likely a good thing to have in any case. Maybe do it as a separate PR?

include/proteus/CompilationTask.hpp
```diff
@@ -0,0 +1,141 @@
...
+#include "proteus/Hashing.hpp"
+#include "proteus/JitCache.hpp"
+#include "proteus/JitStorageCache.hpp"
+#include "proteus/Utils.h"
+
+namespace proteus {
+
+using namespace llvm;
+
+class CompilationTask {
```
davidbeckingsale, line 20: What a great abstraction :)

ggeorgakoudis:
> Looks good! My only comment would be that CompilerSync and CompilerAsync should have the same API (e.g. for the compile method). Why not just have compile always return a MemoryBuffer, and move the takeCompilationResult logic inside of the Async compile?

That's a good suggestion but we'll need to redesign a bit more. The `compile` interface takes a `CompilationTask` which performs an "expensive" clone at construction plus it copies some data containers. I want to avoid constructing it when compilation is pending for the async case, that's why the "split" interface. I'll note this for a later iteration.

PR #169, opened 2025-03-04
Implement lightweight clone for kernel module

davidbeckingsale:


include/proteus/CoreLLVMDevice.hpp
```diff
@@ -234,6 +237,129 @@ inline void specializeIR(Module &M, StringRef FnName, StringRef Suffix,
   runCleanupPassPipeline(M);
 }
 
+inline std::unique_ptr<Module> cloneKernelFromModule(Module& M, const std::string& name) {
+      auto KernelModule = std::make_unique<Module>("JitModule", M.getContext());
```
koparasy, line 241: I believe it would be more clean if you passed a context as a parameter in the function. To enable the user to decouple the lifetime of the generated module with the lifetime of source module. You can pass the   [proteus-context](https://github.com/Olympus-HPC/proteus/blob/331caa4ed7fb49510693a53058e888ef918b064c/include/proteus/JitEngineDevice.hpp#L359) when you call this function. 

What you do now is correct, as you pass the BinModule and that has the lifetime of the entire simulation. But this header file is also part of the CoreProteus target. So you can enable the user to do that.

include/proteus/CoreLLVMDevice.hpp
```diff
@@ -234,6 +237,129 @@ inline void specializeIR(Module &M, StringRef FnName, StringRef Suffix,
...
+      KernelModule->setSourceFileName(M.getSourceFileName());
+      KernelModule->setDataLayout(M.getDataLayout());
+      KernelModule->setTargetTriple(M.getTargetTriple());
+      KernelModule->setModuleInlineAsm(M.getModuleInlineAsm());
+      KernelModule->IsNewDbgInfoFormat = M.IsNewDbgInfoFormat;
+
+      auto KernelFunction =
+          M.getFunction(name);
```
koparasy, line 249: Add an assertion here?

include/proteus/CoreLLVMDevice.hpp
```diff
@@ -234,6 +237,129 @@ inline void specializeIR(Module &M, StringRef FnName, StringRef Suffix,
...
+      ValueToValueMapTy VMap;
+
+      for (auto GV : ReachableGlobals) {
+        GlobalVariable *NewGV = new GlobalVariable(
+            *KernelModule, GV->getValueType(), GV->isConstant(),
+            GV->getLinkage(),
+            GV->hasInitializer() ? GV->getInitializer() : nullptr,
+            GV->getName(),
+            nullptr,
+            llvm::GlobalValue::NotThreadLocal,
```
koparasy, line 314: Isn't there some way to copy this value from the src Global?

include/proteus/CoreLLVMDevice.hpp
```diff
@@ -234,6 +237,129 @@ inline void specializeIR(Module &M, StringRef FnName, StringRef Suffix,
...
+        for (const User *Usr : GV.users()) {
+          const Instruction *I = dyn_cast<Instruction>(Usr);
+
+          if (I) {
+            process_instruction(GV, I);
+          } else {
+            for (const User *NextUser : Usr->users()) {
+              I = dyn_cast<Instruction>(NextUser);
+              if (I)
+                process_instruction(GV, I);
```
ggeorgakoudis, line 298: Where does this need for second-level user iteration come from? Is it because of casting? We should generalize with `Worklist` but I'm OK to do this in a refinement iteration, but please add comment

ggeorgakoudis (commented):
LGTM. 

Address points:
- Dinos's comments
- Linter comments that I follow up
- Fix for testing

ggeorgakoudis:
Rebase (mostly) ☝️

include/proteus/CoreLLVMDevice.hpp
```diff
@@ -229,6 +232,174 @@ inline void specializeIR(
...
+
+      NewNamedMD->addOperand(MapMetadata(MDEntry, VMap));
+    }
+  }
+
+#if PROTEUS_ENABLE_DEBUG
+  if (verifyModule(*KernelModule, &errs()))
+    PROTEUS_FATAL_ERROR("Broken mini-module found, JIT compilation aborted!");
+
+  Logger::logs("proteus") << "Mini-module \n" << *KernelModule << "\n";
```
davidbeckingsale, line 397: Probably want to put the log above the verify - since if we get a fatal error we will never see the module.

include/proteus/CoreLLVMDevice.hpp
```diff
@@ -229,6 +232,174 @@ inline void specializeIR(
   runCleanupPassPipeline(M);
 }
 
+inline std::unique_ptr<Module> cloneKernelFromModule(Module &M, LLVMContext &C,
+                                                     const std::string &Name) {
+#if PROTEUS_ENABLE_DEBUG
+  Logger::logs("proteus") << "Full module \n" << M << "\n";
```
davidbeckingsale, line 238: I think this should be removed - risk of logging something absolutely huge.

davidbeckingsale (commented):
Looks good to me!

ggeorgakoudis (approved):

PR #175, opened 2025-03-12
Do not propagate static library linkage

tbennun:
This PR is necessary for projects with multiple dependent targets, in order for it not link with LLVM more than once.


ggeorgakoudis (approved):
LGTM.

Let's merge when (if?) tests pass

PR #178, opened 2025-03-14
Support JIT GPU static shared arrays

ggeorgakoudis:


include/proteus/JitInterface.hpp
```diff
@@ -44,6 +44,14 @@ static __attribute__((noinline)) T &&register_lambda(T &&t,
   return std::forward<T>(t);
 }
 
+#if defined(__CUDACC__) || defined(__HIP__)
+template <typename T>
+__device__ __attribute__((noinline)) T *
+shared_array(size_t N, size_t ElemSize = sizeof(T)) {
+  __builtin_trap();
```
koparasy, line 51: I would think that compilers may warn about function that does not return something.

adding a return nullptr; would fix this.

include/proteus/TransformSharedArray.hpp
```diff
@@ -0,0 +1,105 @@
...
+          ArrayType *AType =
+              ArrayType::get(Type::getInt8Ty(M.getContext()), N * Sizeof);
+          constexpr unsigned SharedMemAddrSpace = 3;
+          GlobalVariable *SharedMemGV = new GlobalVariable(
+              M, AType, false, GlobalValue::InternalLinkage,
+              UndefValue::get(AType), ".proteus.shared", nullptr,
+              llvm::GlobalValue::NotThreadLocal, SharedMemAddrSpace, false);
+          // Using 16-byte alignment based on AOT code generation.
+          // TODO: Create or find an API to query the proper ABI alignment.
+          SharedMemGV->setAlignment(Align{16});
```
koparasy, line 61: Maybe you can use (this)[https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/IR/DataLayout.h#L320]

include/proteus/JitInterface.hpp
ggeorgakoudis: The builtin trap is `noreturn`, so no compiler complains

tomstitt (approved):
Thanks @ggeorgakoudis !

PR #181, opened 2025-03-15
Fix lambda specialization for caching and assumptions

### LINKED ISSUE #184: Support more complex JIT lambda interactions aka. remove the one pending lambda assumption
The "one pending lambda" assumption has the benefit of simplifying handling lambdas: pushed JIT variables are consumed at only once and caching just pulls the JIT captured variables vector (if there is one). However, this creates a number of issues:

1. One definition, multiple calls
```cpp
auto Lambda = proteus::register_lambda([=, x=proteus::jit_variable(x)]() {...});

launcher(Lambda);
launcher(Lambda);
```
The second launcher will not specialize the lambda since the pushed JIT variables will be consumed at the first launcher call. 

2. `__host__` `__device__` functions issues
```cpp
launcher(proteus::register_lambda(
  [=, x=proteus::jit_variable(x) __host__ __device__ () 
        __attribute__((annotate("jit"))) {...}
  ));
```
If there are invocations of both the `__host__` and `__device__` instantiation of the lambda, the first one will consume the pushed JIT variable and others will not specialize.

Issues 1 and 2 are "easily" solvable by using a map from the lambda symbol to a list of JIT variables (we implement that).

3. Hashing for caching is challenging

For host lambdas, we directly JIT the lambda function itself, so including the pushed JIT variables in the hashing calculation is straightforward.

For device lambda, we JIT the _kernel_ module that calls the lambda and perform the transformation on that module. This mean that at `compileAndRun` we don't have information for which lambda JIT variables we should hash to uniquely identify this kernel specialization.

Possible approaches with their problems:
1. Use all pushed JIT variables for hashing: false misses in the cache when an unrelated lambda is included in the hash (also for host JIT code)
2. Find called lambdas for a kernel to include only their pushed JIT variables when hashing: needs analysis at the pass which may be undecidable (e.g., a function calls the lambda through a pointer)
3. Lazily at runtime find called lambdas for a kernel to include only their pushed JIT variables when hashing: find the first time a kernel is JIT compiled all JIT lambdas callees, record those callees `JITKernelInfo`, retrieve their JIT variables when caching.

Approach 3 sounds best

ggeorgakoudis:
- Include lambda JIT variables when hashing
- Support multiple uses of the same lambda definition
- Support JIT for `__host__` `__device__` lambdas
- Add tests

Closes #184


include/proteus/LambdaRegistry.hpp
```diff
@@ -7,21 +7,28 @@
...
-#include "proteus/Hashing.hpp"
+#include "proteus/Error.h"
 #include "proteus/Logger.hpp"
 
 namespace proteus {
 
 using namespace llvm;
 
+// The LambdaRegistry stores the unique lambda type symbol and JIT member
+// variables. We assume there is always ONLY ONE pending lambda registration
```
davidbeckingsale, line 18: Why are we making this assumption?

ggeorgakoudis: Because we clear(ed) pending JIT variables when `registerLambda` executes, so pushed JIT variables must correspond only in the immediate following`registerLambda`. This update here fixes caching to include lambda captures and codifies that assumption. We can improve in the future but we'll need to discuss.

ggeorgakoudis:
Rebase ☝️

davidbeckingsale (approved):

PR #199, opened 2025-04-16
Add logfile method to Logger to store modules

ggeorgakoudis:
- Move modules to files instead of the output stream


davidbeckingsale (approved):
Nice!

PR #206, opened 2025-04-30
Update logging for console output and timers

ggeorgakoudis:
This PR adds logging to console with timers on the components of compiling a kernel.

### DIFF
--- a/include/proteus/Config.hpp
+++ b/include/proteus/Config.hpp
@@ -1,0 +1,69 @@
+#ifndef PROTEUS_CONFIG_HPP
+#define PROTEUS_CONFIG_HPP
+
+#include <string>
+
+namespace proteus {
+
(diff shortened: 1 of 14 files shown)


davidbeckingsale:
I think it would be nice to have the timers independent of all the other debug logging.

ggeorgakoudis:
> I think it would be nice to have the timers independent of all the other debug logging.

Controlled by a runtime env var instead of the PROTEUS_ENABLE_DEBUG guard?

davidbeckingsale:
> Controlled by a runtime env var instead of the PROTEUS_ENABLE_DEBUG guard?

Perfect!

ggeorgakoudis:
Also rebased ☝️

ggeorgakoudis:
Rebase ☝️

PR #216, opened 2025-05-15
Move things around

koparasy:
Observations:
- [ ] [SetKernelDims](https://github.com/Olympus-HPC/proteus/blob/20168b1d4f5021f9bb56403d070a9b16a63df977/include/proteus/CoreLLVMDevice.hpp#L30) is a module level modification. They way we use is fine, as we use it after cloning. Maybe though we need to have some way to clarify this.


koparasy:
@ggeorgakoudis I will keep this branch also for some notes on functionality.

PR #218, opened 2025-05-15
Support HIP compressed offload bundles

ggeorgakoudis:


ggeorgakoudis:
> @ggeorgakoudis Thanks!! Does this mean that compressed bundles would JIT slower due to decompression on every call, or is that optimized w.r.t. the JIT cache?

Proteus will cache extracted modules after decompression so it won't decompress on every invocation. It shouldn't be an issue but good to keep an eye on it.

tbennun (approved):
LGTM!

PR #235, opened 2025-07-30
Support object specialization

ggeorgakoudis:
- Introduce the proteus::jit_object inteface to specialize for trivially copyable objects
- Detect argument coercion (scalars, vectors) for JIT objects and make those coerced arguments as runtime constants and improve analysis
- Support runtime constant info for objects, also static arrays and vectors since they are used for argument coercion
- Update push JIT variable for lambdas to type the runtime constant
- Use the runtime constant type as the authoritative source for argument specialization
- Simplify JIT cache insertion interface to avoid copying the runtime constant vector
- Add tests

### DIFF
--- a/include/proteus/JitEngineDevice.hpp
+++ b/include/proteus/JitEngineDevice.hpp
@@ -617,7 +617,7 @@ JitEngineDevice<ImplT>::compileAndRun(
       auto KernelFunc =
           getKernelFunctionFromImage(KernelMangled, CacheBuf->getBufferStart());
 
-      CodeCache.insert(HashValue, KernelFunc, KernelInfo.getName(), RCVec);
+      CodeCache.insert(HashValue, KernelFunc, KernelInfo.getName());
 
       return launchKernelFunction(KernelFunc, GridDim, BlockDim, KernelArgs,
                                   ShmemSize, Stream);
@@ -679,7 +679,7 @@ JitEngineDevice<ImplT>::compileAndRun(
       KernelMangled, ObjBuf->getBufferStart(),
       Config::get().ProteusRelinkGlobalsByCopy, VarNameToDevPtr);
 
-  CodeCache.insert(HashValue, KernelFunc, KernelInfo.getName(), RCVec);
+  CodeCache.insert(HashValue, KernelFunc, KernelInfo.getName());
   if (Config::get().ProteusUseStoredCache) {
     StorageCache.store(HashValue, ObjBuf->getMemBufferRef());
   }
--- a/include/proteus/JitInterface.hpp
+++ b/include/proteus/JitInterface.hpp
@@ -52,18 +52,59 @@ jit_array(T V, [[maybe_unused]] size_t NumElts,
 #endif
 
 template <typename T>
-static __attribute__((noinline)) T jit_variable(T v, int pos = -1) {
-  RuntimeConstant RC;
-  std::memcpy(&RC, &v, sizeof(T));
-  RC.Slot = pos;
-  __jit_push_variable(RC);
+__attribute__((noinline)) void
+jit_object(T *V, size_t Size = sizeof(std::remove_pointer_t<T>)) noexcept;
 
-  return v;
+#if defined(__CUDACC__) || defined(__HIP__)
+template <typename T>
+__attribute__((noinline)) __device__ void
+jit_object(T *V, size_t Size = sizeof(T)) noexcept;
+#endif
+
+template <typename T>
+__attribute__((noinline)) void
+jit_object(T &V, size_t Size = sizeof(std::remove_reference_t<T>)) noexcept;
+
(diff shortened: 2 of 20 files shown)


include/proteus/JitInterface.hpp
```diff
@@ -52,18 +52,59 @@ jit_array(T V, [[maybe_unused]] size_t NumElts,
...
+
+template <typename T>
+__attribute__((noinline)) void
+jit_object(T &V, size_t Size = sizeof(std::remove_reference_t<T>)) noexcept;
+
+#if defined(__CUDACC__) || defined(__HIP__)
+template <typename T>
+__attribute__((noinline)) __device__ void
+jit_object(T &V, size_t Size = sizeof(T)) noexcept;
+#endif
```
koparasy, line 72: Should we have static checks or `std::enable_if<std::is_trivially_copyable<T>::value>::type` across these?

include/proteus/JitInterface.hpp
```diff
@@ -52,18 +52,59 @@ jit_array(T V, [[maybe_unused]] size_t NumElts,
...
 
-  return v;
+  return V;
 }
 
 template <typename T>
-static __attribute__((noinline)) T &&register_lambda(T &&t,
-                                                     const char *Symbol = "") {
+static __attribute__((noinline)) T &&
+register_lambda(T &&t, const char *Symbol = "") noexcept {
```
koparasy, line 107: What does noexcept do? Improve code gen for your analysis?

src/pass/ProteusPass.cpp
```diff
@@ -758,8 +817,17 @@ class ProteusPassImpl {
       for (size_t ArgI = 0; ArgI < StubFn->arg_size(); ++ArgI) {
         auto *GEP = Builder.CreateInBoundsGEP(
             ArgPtrsTy, ArgPtrs, {Builder.getInt32(0), Builder.getInt32(ArgI)});
-        Builder.CreateStore(StubFn->getArg(ArgI), ArgPtrAllocas[ArgI]);
-        Builder.CreateStore(ArgPtrAllocas[ArgI], GEP);
+        // If the argument has the 'byval' or 'byref' attribute, we store the
+        // pointer to the argument value directly in the alloca, otherwise we
+        // store the pointer-to-pointer. This is done to conform to the ABI and
```
koparasy, line 822: This is annoying. I guess you reversed engineer this. Is there some assertion we can have to put here in the case some other ABI/Arch handles the byval/ref differently?

tests/cpu/jit_struct.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+struct DimMix {
+  int X;
+  double Y;
+  float Z;
+};
+
+struct DimFill {
+  int X, Y, Z;
+  int Fill[16];
```
koparasy, line 31: Here you have "16" on the gpu you have 10. I have a small preference towards 10. I would likely make it 7. Something really weird to push paddings in it.

koparasy (approved):
I have some nits here and there.

include/proteus/JitInterface.hpp
ggeorgakoudis: Simplifies codegen and really we don't throw exceptions

src/pass/ProteusPass.cpp
ggeorgakoudis: I can't think of something :/. Both in LLVM IR and runtime it looks like it's "just" a pointer but the semantics depend on the ABI conventions. Our detection is to use the byval/byref attributes.

tests/cpu/jit_struct.cpp
ggeorgakoudis: For GPUs it doesn't matter, the object always passes as a pointer. Any number of elements will align it on a 4-byte boundary (base type is `int`). I can try a `char Fill[7]` or something, I'll make it a todo.

include/proteus/JitInterface.hpp
koparasy: sg

include/proteus/JitEngineDevice.hpp
```diff
@@ -617,7 +617,7 @@ JitEngineDevice<ImplT>::compileAndRun(
       auto KernelFunc =
           getKernelFunctionFromImage(KernelMangled, CacheBuf->getBufferStart());
 
-      CodeCache.insert(HashValue, KernelFunc, KernelInfo.getName(), RCVec);
+      CodeCache.insert(HashValue, KernelFunc, KernelInfo.getName());
```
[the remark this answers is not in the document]
koparasy: Sounds good. Made an issue for this.

src/pass/ProteusPass.cpp
koparasy: Ok. We already test x86, power9. So we cover the architectures we use. If we find an ARM free CI would be great to test it there as well.

PR #245, opened 2025-08-12
Clean up launch bound setting

koparasy:


ggeorgakoudis:
I'm not fond of changing the interface of `setLaunchBoundsForKernel` to take target specific arguments. I'd expect we pass the `BlockSize`, `GridSize`, which we want to know at runtime, and let the target specific functions handle it to setup launch bounds. Can you explain the rationale? Are you using this to tune target-specific LB parameters from Mnene?

koparasy:
> I'm not fond of changing the interface of `setLaunchBoundsForKernel` to take target specific arguments. I'd expect we pass the `BlockSize`, `GridSize`, which is what we know at runtime, and let the target specific functions handle it to setup launch bounds. Can you explain the rationale? Are you using this to tune target-specific LB parameters from Mnene?

Since these functions are defined under Core*.hpp, I think they should provide the fundamental capability to set attribute bounds directly as defined by the vendor.  This ensures that we have a consistent, low-level interface, while allowing vendor-specific implementations at the LLVM level to interpret those parameters according to their own semantics. For example, mneme can use that capability and tune these parameters. 

If, in the future, we find that a runtime heuristic is valuable, we can incorporate it cleanly by calling something that looks like:

```cpp
setLaunchBoundsForKernel(F, BlockSize, device::getMinBlocks(Blocks, Grid, IR));
```

Here, `getMinBlocks` from the JitEngine function would handle the runtime logic, ensuring that any heuristic remains separate from the core API.

ggeorgakoudis:
Makes sense, I see the point. I'll review in more detail.

include/proteus/CoreLLVMCUDA.hpp
```diff
@@ -84,36 +84,40 @@ inline const SmallVector<StringRef> &threadIdxZFnName() {
 
 } // namespace detail
 
-inline void setLaunchBoundsForKernel(Module &M, Function &F,
-                                     size_t /*GridSize*/, int BlockSize) {
-  NamedMDNode *NvvmAnnotations = M.getNamedMetadata("nvvm.annotations");
+inline void setLaunchBoundsForKernel(Function &F, int MaxThreadsPerSM,
+                                     int MinBlocksPerSM = 0) {
+  auto M = F.getParent();
```
ggeorgakoudis, line 89: Fix linter comment: `auto *M`

include/proteus/CoreLLVMCUDA.hpp
```diff
@@ -84,36 +84,40 @@ inline const SmallVector<StringRef> &threadIdxZFnName() {
...
-    // Expecting 3 operands ptr, desc, i32 value.
-    assert(MetadataNode->getNumOperands() == 3);
-
-    auto *PtrMetadata = MetadataNode->getOperand(0).get();
-    auto *DescMetadata = MetadataNode->getOperand(1).get();
-    if (PtrMetadata == FuncMetadata && MaxntidxMetadata == DescMetadata) {
-      MetadataNode->replaceOperandWith(2, MaxThreadsMetadata);
-      return;
+
+  auto setMDNode = [&](const char *MDName, int MDValue) {
```
ggeorgakoudis, line 94: Fix linter comment: `SetMDNode`

include/proteus/CoreLLVMHIP.hpp
```diff
@@ -513,23 +513,26 @@ inline std::unique_ptr<MemoryBuffer> codegenRTC(Module &M,
...
+    // For more information please check:
+    // https://clang.llvm.org/docs/AttributeReference.html#amdgpu-waves-per-eu
+    F.addFnAttr("amdgpu-waves-per-eu",
+                std::to_string(WavesPerEU) + "," + std::to_string(WavesPerEU));
+  }
+
   PROTEUS_DBG(Logger::logs("proteus")
-              << "BlockSize " << BlockSize << " GridSize " << GridSize
-              << " => Set Wokgroup size " << BlockSize
+              << " => Set Wokgroup size " << MaxNumWorkGroups
```
ggeorgakoudis, line 535: Fix preexisting typo: `Workgroup`

ggeorgakoudis (changes requested):
LGTM

Fix linter and typo comments

ggeorgakoudis (approved):
LGTM

PR #247, opened 2025-08-12
Frontend GPU: Implement FloydWarshall Benchmark

ZwFink:
Ported FloydWarshall from HeCBench HIP. Verified that it produces the same results. To support the benchmark, I added `unsigned int` to the frontend TypeMap/Var.
Note that I flattened the 2D grid because current `builtins` only support 1D.
The JIT version is about 2% faster than the default, but it may be a worthwhile addition because it extends required support to `unsigned int`.


tests/frontend/gpu/floydwarshall.cpp
```diff
@@ -0,0 +1,330 @@
+// NOLINTBEGIN
+
+// clang-format off
+// RUN: rm -rf .proteus
+// RUN: ./floydwarshall.%ext 1024 100 16 | %FILECHECK %s --check-prefixes=CHECK,CHECK-FIRST
```
ggeorgakoudis, line 5: Add `PROTEUS_TRACE_OUTPUT=1` and check that we get the specialization printout

tests/frontend/gpu/floydwarshall.cpp
```diff
@@ -0,0 +1,330 @@
...
+// CHECK-NEXT: dist[1048574] = 4
+// CHECK-NEXT: dist[1048575] = 0
+// CHECK: JitCache hits 102399 total 102400
+// CHECK: HashValue {{[0-9]+}} NumExecs 102400 NumHits 102399
+// CHECK-FIRST: JitStorageCache hits 0 total 1
+// CHECK-SECOND: JitStorageCache hits 1 total 1
+// CHECK: JitCache hits 0 total 0
+// CHECK: JitStorageCache hits 0 total 0
+
+// NOLINTEND
```
ggeorgakoudis, line 330: End with a newline

ggeorgakoudis (commented):
LGTM

Few minor changes.

koparasy:
@ggeorgakoudis should we move most of these code (and adam) under the proteus-benchmark? Not in this PR but in another one.

koparasy:
> > @ggeorgakoudis should we move most of these code (and adam) under the proteus-benchmark? Not in this PR but in another one.
> 
> A third option is to add an `examples` directory in Proteus to give people checking the project out a way to explore Proteus in less-trivial scenarios than, say, daxpy, without having to clone another repository, get the build working, etc.

This is an interesting direction and probably I favor it as proteus benchmarks is more complex than the entry point user. Yet, we need to be aware of licensing. 

I am not a licensing person but can we copy Hecbench source to examples?

tests/frontend/gpu/floydwarshall.cpp
ZwFink: It looks like the frontend doesn't go through the specialization. It goes `Dispatch.compile` -> (dispatcher) `Jit.compileOnly`.

ggeorgakoudis:
> @ggeorgakoudis should we move most of these code (and adam) under the proteus-benchmark? Not in this PR but in another one.

Yes, I set a bad precedent with adam 😁

ggeorgakoudis:
> > @ggeorgakoudis should we move most of these code (and adam) under the proteus-benchmark? Not in this PR but in another one.
> 
> A third option is to add an `examples` directory in Proteus to give people checking the project out a way to explore Proteus in less-trivial scenarios than, say, daxpy, without having to clone another repository, get the build working, etc.

The examples directory is a great idea but I think we need less complicated programs than floydwarshall.

PR #255, opened 2025-08-21
bug: fix segfault caused by long filename

johnbowen42:


koparasy (commented):
The changes themselves look good overall. However, they introduce breaking changes to the core interface, which we intended to keep header-only and lightweight.

The main concern isn’t the modified function signatures, but rather the inclusion of Hashing.h. That header pulls in additional Proteus dependencies and, potentially in the future, may also introduce function declarations without definitions. This would break the principle of having a self-contained, header-only core.

The current integration tests IIRC do not test specialize at all and thus they may not raise such linking issues.

Whether we should revisit or not this approach to ensure the core remains minimal and dependency-free is your decision @ggeorgakoudis  and @johnbowen42 .

PR #259, opened 2025-08-27
[cpp-frontend] Support modules with host plus device code

ggeorgakoudis:
- Add HOST_HIP, HOST_CUDA target models
- Add HostHIP and HostCUDA dispatchers
- Add compileCppToDynamicLibrary method that uses the clang driver to compile mixed, host plus device modules
- Update host jit engine to load dynamic libraries produced by the cpp jit module for symbol lookup
- Add test


ZwFink:
I think you have have forgotten to commit the file `tests/frontend/host_device/CMakeLists.txt`

ZwFink (approved):

PR #260, opened 2025-08-27
performance: fix inlining regression for CUDA backends

johnbowen42:
This PR fixes regressions in ENERGY and FIRST_MIN raja perf benchmarks for the CUDA backend.  It does so by removing `MergeFunctionsPass` from the Proteus plugin pass.  Previously, this pass was calling instantiations of `template<typename T> thread_privatize` with mismatching return types, breaking the inliner.  

This PR has been evaluated on both NVIDIA and HIP backends.  The regressions for ENERGY and FIRST_MIN are solved, and performance portability is maintained for other RAJAPerf benchmarks and backends.  

I also introduce a RAJA style launch unit test, which was useful for debugging the inlining regression. 

We also gain one speedup on the RAJAPerf suite, FIRST_MIN (previously a slowdown)


johnbowen42:
NVIDIA benchmarks, Lassen
nvidia-bar-kernel-speedup-ablation-0
nvidia-bar-kernel-speedup-ablation-1
nvidia-bar-kernel-speedup-ablation-2
nvidia-bar-kernel-speedup-ablation-3

tests/gpu/raja_style_launch.hpp
ggeorgakoudis, line 1: Turn linter off for this file to avoid the warnings

tests/gpu/inlining_regression.cpp
```diff
@@ -0,0 +1,55 @@
...
+  gpuErrCheck(gpuFree(Y));
+
+  proteus::finalize();
+}
+
+// clang-format off
+// CHECK: 0
+// CHECK-FIRST: [ArgSpec] Replaced Function _Z9daxpyImpldPdS_m ArgNo 3 with value i64 1024
+// CHECK-FIRST: [DimSpec]
+// CHECK-FIRST: [DimSpec]
```
ggeorgakoudis, line 49: Check for values here

ggeorgakoudis (commented):
LGTM, few comments

ggeorgakoudis:
Thanks @johnbowen42! The plots are very useful. Can you clarify these two points:

1. Do we have any performance regressions for HIP programs after this change?
2. Do we have any performance regressions for other CUDA programs after this change?

johnbowen42:
> Thanks @johnbowen42! The plots are very useful. Can you clarify these two points:
> 
> 1. Do we have any performance regressions for HIP programs after this change?
> 2. Do we have any performance regressions for others than the ones it "fixes" CUDA programs after this change?

1. No, HIP performance is unchanged
2. No, the only CUDA programs affected are ENERGY and FIRST_MIN

johnbowen42:
@ggeorgakoudis Do you have any recommendations for checking the IR generated from `inlining_regression.cpp` to make sure inlining is happening?

tests/gpu/inlining_regression.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+  gpuErrCheck(gpuFree(X));
+  gpuErrCheck(gpuFree(Y));
+
+  proteus::finalize();
+}
+
+// clang-format off
+// CHECK: 0
+// CHECK-NOT: thread_privatize
```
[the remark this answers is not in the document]
johnbowen42: We emit a module for each kernel launched--see https://github.com/Olympus-HPC/proteus/blob/7709bb126f742ad86d180b512288d9662d6a1ed9/include/proteus/CompilationTask.hpp#L121

Before this fix, the thread privatize function would correctly inline for just one of the kernel launches (where there was no type mismatch at the call site).  We want to make sure it's inlined for all the kernel launches.

PR #262, opened 2025-08-27
PJ-DSL: Add Array Vars

ZwFink:
- Adds `declVar<T[]>` to `Func` to declare array variables. Arrays can be declared for a specific address space.
- `Var` now can now be created from indexing into arrays. Var uses the concept of borrowed storage for this
  - Originally each `Var` was required to be associated with an `alloca`. This makes less sense for `Var`s that index into arrays.
- Adds SyncThreads intrinsic for GPU code


ZwFink:
@ggeorgakoudis This is ready for review

src/lib/Frontend/Var.cpp
```diff
@@ -66,63 +69,40 @@ static Var &cmpOp(const Var &L, const Var &R, IntOp IOp, FPOp FOp) {
...
 
-Value *Var::getValue() const {
-  auto &IRB = Fn.getIRBuilder();
-  Type *AllocaType = Alloca->getAllocatedType();
-  if (AllocaType->isPointerTy()) {
-    auto *Ptr = IRB.CreateLoad(AllocaType, Alloca);
-    return IRB.CreateLoad(PointerElemType, Ptr);
-  }
-  return IRB.CreateLoad(AllocaType, Alloca);
+Var::Var(Value * /*PointerValue*/, FuncBase &Fn) : Alloca(nullptr), Fn(Fn) {}
```
ggeorgakoudis, line 74: Where is this constructor used?

src/lib/Frontend/Var.cpp
```diff
@@ -66,63 +69,40 @@ static Var &cmpOp(const Var &L, const Var &R, IntOp IOp, FPOp FOp) {
...
+
+StringRef Var::getName() const {
+  PROTEUS_FATAL_ERROR("Invalid Var::getName()");
 }
```
ggeorgakoudis, line 78: Why don't we make invalid functions pure virtual instead of erroring out?

src/lib/Frontend/Var.cpp
ZwFink: > Where is this constructor used?

It's used by the [ArrayVar constructor](https://github.com/Olympus-HPC/proteus/blob/feature/frontend-array/src/lib/Frontend/Var.cpp#L839)

PR #263, opened 2025-08-27
PJ-DSL: Loop Tiling

ZwFink:
Allows to create and tile loop nests.
- Adds `forLoop` for delayed emission of for loops after transformations. Currently, only tiling is supported.
  - The same uniform tilesize can be applied to loop nests, or loops can be tiled individually.
- Adds `buildLoopNest` to tile nested loops.
Here is an example tiled loop nest:
```c++
      F.buildLoopNest(
           F.forLoop({I, Zero, UbnI, IncOne}).tile(TileSize),
            F.forLoop({J, Zero, UbnJ, IncOne}).tile(TileSize),
            F.forLoop({K, Zero, UbnK, IncOne},
                                   [&]() {
                                     auto CIdx = I * N + J;
                                     auto AIdx = I * N + K;
                                     auto BIdx = K * N + J;
                                     C[CIdx] += A[AIdx] * B[BIdx];
                                   })
                .tile(TileSize))
          .emit();
```

- Adds `min` to Var, required to clamp tile size


include/proteus/Frontend/Func.hpp
```diff
@@ -151,6 +153,13 @@ class FuncBase {
     return Lower(*this);
   }
 
+  ForLoopBuilder forLoop(LoopBoundsDescription Bounds);
+  ForLoopBuilder forLoop(LoopBoundsDescription Bounds,
+                         std::function<void()> Body);
```
ggeorgakoudis, line 158: I'm not fond of the `std::function` and its overhead. I'd suggest to make this a templated Lambda and move to implementation to the header:
```
struct EmptyLambda {
    void operator()() const {}
};

template<typename Lambda>
ForLoopBuilder forLoop(LoopBoundsDescription Bounds,
                         Lambda &&Body = {});
```

include/proteus/Frontend/Func.hpp
```diff
@@ -151,6 +153,13 @@ class FuncBase {
...
+
+  LoopNestBuilder buildLoopNest(std::vector<ForLoopBuilder> Loops);
```
ggeorgakoudis, line 160: Instead of a vector I'd use a variadic template to gather the `ForLoopBuilder` templated on lambda objects.

Ping me if that doesn't make sense. 

We may do such refactoring at a later time but it makes time to do it now that you have the code fresh in your mind

include/proteus/Frontend/Func.hpp
```diff
@@ -151,6 +157,16 @@ class FuncBase {
     return Lower(*this);
   }
 
+  template <typename BodyLambda = EmptyLambda>
+  auto forLoop(const LoopBoundInfo &Bounds, BodyLambda &&Body = {}) {
+    return ForLoopBuilder(Bounds, *this, std::move(Body));
+  }
+
+  template <typename... LoopBuilders>
+  auto buildLoopNest(LoopBuilders... Loops) {
```
ggeorgakoudis, line 166: Forwarding reference? `LoopBuilders... &&Loops`?

include/proteus/Frontend/Func.hpp
```diff
@@ -151,6 +157,16 @@ class FuncBase {
...
+    return LoopNestBuilder(*this, std::move(Loops)...);
```
ggeorgakoudis, line 167: `std::forward`?

PR #264, opened 2025-08-28
[cpp-frontend] Make function, kernel handles public

ggeorgakoudis:
- Allows the user to store and re-use


ggeorgakoudis:
Rebase on `main` ☝️

ZwFink (approved):

PR #266, opened 2025-08-28
Improve linking with proteus

ggeorgakoudis:
- Enclose LLVM and Clang libs in a linking group
- Add integration test with GCC AOT compiler using the proteusFrontend


ggeorgakoudis:
Rebase on latest `main` ☝️

PR #269, opened 2025-09-03
Fix linking bug with proteus and non-proteus compiled objects

ggeorgakoudis:
- Do not change linkage and visibility in the original module's symbols
- Remove duplicate check for unresolved symbols in the LLVM IR
- Link consumer tests with `-rdynamic` (as we document)

### DIFF
--- a/tests/cpu/CMakeLists.txt
+++ b/tests/cpu/CMakeLists.txt
@@ -22,6 +22,10 @@ function(CREATE_CPU_TEST exe check_source)
         ${NATIVE_OPT_FLAGS}
     )
 
+    target_link_options(${exe} PRIVATE
+        $<LINK_ONLY:-rdynamic>
+    )
+
     add_test(NAME ${exe} COMMAND ${LIT} -vv -D FILECHECK=${FILECHECK} ${check_source})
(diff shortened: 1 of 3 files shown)


tests/cpu/CMakeLists.txt
```diff
@@ -22,6 +22,10 @@ function(CREATE_CPU_TEST exe check_source)
         ${NATIVE_OPT_FLAGS}
     )
 
+    target_link_options(${exe} PRIVATE
+        $<LINK_ONLY:-rdynamic>
+    )
+
```
koparasy, line 28: I am wondering whether you need to expose this also to you cmake interface for your external users.

ggeorgakoudis: I'm wondering the same but all the reading I do says it's a no-no and that you should leave this option to consumers

koparasy: I guess, in the case of HOST jitting, if the consumer does not use `rdynamic` proteus will be unable to relink the global symbols.

koparasy (approved):

ZwFink (approved):

PR #275, opened 2025-09-05
Remove fixed unrolling factor from rtc opt pipeline

koparasy:
We observed significant performance differences on Bezier when using a Mneme-suggested configuration across RTC and serial code generation. After investigation, we traced the issue to the custom flags passed to RTC. This PR removes those flags to ensure consistent performance.


ggeorgakoudis:
Artifacts PR 275 commit 7d861e8f
<p>








</p>

PR #277, opened 2025-09-05
Fix Compiler Error in LoopNest.tile

ZwFink:
The included test calls `LoopNest.tile`, which does not compile in the current implementation.
The current `LoopNest.tile` contains a fold expression over the tuple of loops. But `ForLoops` with empty bodies have the same type (`ForLoopBuilder<EmptyLambda>`), and the `std::get<LoopBuilders>` fails.

This PR fixes the issue by enumerating the tuple and `std::get`'ting using the enumerated indices.


ggeorgakoudis (approved):
LGTM

PR #278, opened 2025-09-05
PJ-DSL: Unify GPU builtins under proteus::builtins::gpu

ZwFink:
Select HIP/CUDA builtins at build time.

Any PJ-DSL users that link with Proteus would have to do the following in their code:
```c++
#define PROTEUS_ENABLE_[HIP|CUDA] 1
#include <proteus/Frontends/Builtins.hpp>
```

This PR builds the correct builtins based on whether CUDA/HIP are targeted, and makes them accessible under the unified namespace `proteus::builtins::gpu`.

Now, only the file include is required.


ggeorgakoudis (approved):
LGTM

ZwFink:
> > Select HIP/CUDA builtins at build time.
> > Any PJ-DSL users that link with Proteus would have to do the following in their code:
> > ```c++
> > #define PROTEUS_ENABLE_[HIP|CUDA] 1
> > #include <proteus/Frontends/Builtins.hpp>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This PR builds the correct builtins based on whether CUDA/HIP are targeted, and makes them accessible under the unified namespace `proteus::builtins::gpu`.
> > Now, only the file include is required.
> 
> Don't we already propagate the definition `PROTEUS_ENABLE_[HIP|CUDA] ` through cmake already? I guess the user will need to define it if using a more primitive build system? Maybe we ought to produce a config header that has those definitions to include in the user-interface header files

I guess this is probably true. I was prompted to do this when implementing in another directory HeCBench codes that use Makefiles.

PR #279, opened 2025-09-15
PJ-DSL: Turn fp-contract on

ZwFink:
Since v14, clang sets the flag [`-ffp-contract=on`](https://releases.llvm.org/14.0.0/tools/clang/docs/ReleaseNotes.html#floating-point-support-in-clang) by default, enabling IR instruction rewrites into FMA.
By setting this flag in PJ-DSL's IRBuilder, we match Clang's default behavior, which the CPP and annotation JIT frontends both use.

This improves PJ-DSL performance, eliminating most of its slowdown relative to the other frontends (remaining slowdown is attributed to no launch bound specialization).

GEMM (HIP, Tuolumne) without fp-contract:
```
Average over 25 trials: 89.928 ms
```

GEMM (HIP, Tuolumne) with fp-contract:
```
Average over 25 trials: 83.211 ms
```

One value in the `adam` tests changes as a result. FMA is generally considered more accurate than the unfused multiply + add because it rounds once, not twice.

This PR directly fixes an acute performance regression. Longer term, we should allow the user to choose fast math flags, with default values matching Clang.

### DIFF
--- a/src/lib/Frontend/Func.cpp
+++ b/src/lib/Frontend/Func.cpp
@@ -7,6 +7,16 @@ FuncBase::FuncBase(JitModule &J, FunctionCallee FC)
   Function *F = cast<Function>(FC.getCallee());
   BasicBlock::Create(F->getContext(), "entry", F);
   Name = F->getName();
+
+  // Clang enables the 'contract' rewrite rule by default to enable FMA
+  // instructions.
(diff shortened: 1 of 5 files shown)


ggeorgakoudis (approved):
LGTM

You have a todo on making codegen options configurable. Leave the todo and add an issue too

PR #281, opened 2025-09-16
performance: Fix remaining CUDA regressions

johnbowen42:
This PR fixes the remaining performance regressions caused by Proteus when integrating with RAJAPerf on the CUDA backend.  The fix is simple: remove MergeFunctionsPass from the cleanup pipeline (this pass can break inlining in LLVM IR generated from heavily templated cpp code like RAJA), and disable `llvm.assume` intrinsics for the CUDA backend


johnbowen42:
NVIDIA plots: measurements on Lassen with CUDA 12.0.0.  Regressions in LTIMES is fixed.  Regressions in TRAP_INT, REDUCE_SUM, REDUCE3_INT, and PI_REDUCE are turned into speedups roughly around 1.2.  Speedup in FIRST_MIN and REDUCE_STRUCT is improved
nvidia-bar-kernel-speedup-ablation-0
nvidia-bar-kernel-speedup-ablation-1
nvidia-bar-kernel-speedup-ablation-2
nvidia-bar-kernel-speedup-ablation-3
nvidia-bar-kernel-speedup-ablation-4
nvidia-bar-kernel-speedup-ablation-5
nvidia-bar-kernel-speedup-ablation-6
nvidia-bar-kernel-speedup-ablation-7
nvidia-bar-kernel-speedup-ablation-8
nvidia-bar-kernel-speedup-ablation-9
nvidia-bar-kernel-speedup-ablation-10
nvidia-bar-kernel-speedup-ablation-11
nvidia-bar-kernel-speedup-ablation-12
nvidia-bar-kernel-speedup-ablation-13
nvidia-bar-kernel-speedup-ablation-14
nvidia-bar-kernel-speedup-ablation-15
nvidia-bar-kernel-speedup-ablation-16

johnbowen42:
Note: we lose speedups of 1.05 in 2MM and 1.10 in GEMM with this change

include/proteus/CoreLLVMDevice.hpp
```diff
@@ -111,6 +111,8 @@ inline void setKernelDims(Module &M, dim3 &GridDim, dim3 &BlockDim) {
     }
   };
 
+// These optimizations cause regressions for CUDA
+#if defined(PROTEUS_ENABLE_HIP)
```
ggeorgakoudis, line 115: This causes errors for unused variable. Better make the call a no-op for CUDA:
```cpp
auto InsertAssume = ... {
// No-op for CUDA because this specialization causes performance regressions.
#if PROTEUS_ENABLE_CUDA
  return;
#endif
...
}
```

ggeorgakoudis:
Is it only the `llvm.assume` that causes the performance regression? Replacing the CUDA getter intrinsic does not have the same effect?

PR #284, opened 2025-09-17
PJ-DSL: Call Functions with Args

ZwFink:
This PR adds function arguments to PJ-DSL for external and internal functions. It also brings PJ-DSL `addFunction`/`addKernel` syntax in line with CPPJitModule.
```c++
auto &F = J->addFunction<void(int)>("function");
...
auto &V3 = F.call<int(int, int)>("add", V1, V2);
```

The function's signature is repeated at callsites so internal/external functions maintain the same interface.

### DIFF
--- a/include/proteus/Frontend/Func.hpp
+++ b/include/proteus/Frontend/Func.hpp
@@ -30,11 +30,12 @@ template <typename RetT_, typename... ArgT> struct FnSig<RetT_(ArgT...)> {
   using RetT = RetT_;
 };
 
-  template <typename RetT, typename... ArgT>
-  static FunctionCallee getFunctionCallee(Module &M, LLVMContext &Ctx, StringRef Name, ArgTypeList<ArgT...>) {
-    return M.getOrInsertFunction(Name, TypeMap<RetT>::get(Ctx),
-                                  TypeMap<ArgT>::get(Ctx)...);
-  }
+template <typename RetT, typename... ArgT>
+static FunctionCallee getFunctionCallee(Module &M, LLVMContext &Ctx,
+                                        StringRef Name, ArgTypeList<ArgT...>) {
+  return M.getOrInsertFunction(Name, TypeMap<RetT>::get(Ctx),
+                               TypeMap<ArgT>::get(Ctx)...);
+}
 
 using namespace llvm;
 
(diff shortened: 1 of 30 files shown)


ZwFink:
> Great addition! We need this!
> 
> The only change I'd ask is to give a real signature in the template (which we can break down internally as needed, see `CppJitFrontend` for the handling of signatures)
> 
> Instead of:
> 
> ```c++
> F.call<double, double, double>(...)
> ```
> 
> do
> 
> ```c++
> F.call<double(double, double)>
> ```

Done.

ZwFink:
> Great addition! We need this!
> 
> The only change I'd ask is to give a real signature in the template (which we can break down internally as needed, see `CppJitFrontend` for the handling of signatures)
> 
> Instead of:
> 
> ```c++
> F.call<double, double, double>(...)
> ```
> 
> do
> 
> ```c++
> F.call<double(double, double)>
> ```

Done (for real this time).
Later, it would be worthwhile to support the following syntax:
```c++
auto &F1Handle = J->addFunction<void(int)>("func");
auto &F = J->addFunction<void(void)>("otherFunc");
F.beginFunction()
...
F.call(F1Handle, int);
```
So we don't have to duplicate the signature at each callsite (we could store it in the function handle itself).
But I think it's fine to keep as-is for now: I have a pressing need for function calls with arguments, the nicer syntax is secondary.

include/proteus/Frontend/Func.hpp
```diff
@@ -21,6 +21,22 @@ class LoopBoundInfo;
...
+// Useful to partially-specialize function templates.
+template <typename... ArgTs> struct ArgTypeList {};
+template <typename T> struct FnSig;
+template <typename RetT_, typename... ArgT> struct FnSig<RetT_(ArgT...)> {
+  using ArgsTList = ArgTypeList<ArgT...>;
+  using RetT = RetT_;
+};
+
+template <typename RetT, typename... ArgT>
+static FunctionCallee getFunctionCallee(Module &M, LLVMContext &Ctx,
```
ggeorgakoudis, line 34: `inline` instead of `static` (no need to internalize at inclusion point), or better yet make `getFunctionCallee` a method of `JitModule` that uses `Module` and `LLVMContext` directly from the class object

ggeorgakoudis (approved):
LGTM

One change

PR #285, opened 2025-09-19
Set LLVM_DIR to real path

koparasy:
Fixes an issue where CMake reports a mismatch between

    /opt/rocm-6.3.1/llvm/lib/cmake/llvm
    /opt/rocm-6.3.1/llvm//lib/cmake/llvm

Although these point to the same directory, the redundant "//"
causes Proteus to treat them as different. This patch normalizes
LLVM_DIR paths before comparison.

Note: /opt/rocm-6.3.1/llvm is a symlink to /opt/rocm-6.3.1/lib/llvm,
so projects that integrate with Proteus (e.g. Mneme) must also
resolve LLVM_DIR consistently.


ggeorgakoudis (approved):
LGTM but by https://cmake.org/cmake/help/latest/command/get_filename_component.html`get_filename_component` is superseded by `file(REAL_PATH)`. Please make the change.

PR #288, opened 2025-09-23
Group all options modifying code generation under a class

koparasy:
This PR lays the groundwork for supporting per-kernel Proteus configurations instead of relying solely on a single global configuration. The long-term goal is to allow fine-grained tuning for each kernel, enabling more precise optimizations.

## Planned Follow-up
An upcoming PR will introduce:

```C++
DenseMap<std::string, CodeGenerationConfig> TuneConfigs;
```

This will map kernel names (and later possibly static/dynamic hashes) to tuned options detected by Mneme.

## Changes in this PR

* Introduces the CodeGenerationConfig class as the core abstraction for kernel-level tuning.
* Initializes and populates environment-based settings in the global CodeGenerationConfig.

PR #289, opened 2025-09-24
Features/per kernel config

koparasy:
Depends on #288


tests/gpu/CMakeLists.txt
```diff
@@ -214,6 +214,14 @@ elseif(PROTEUS_ENABLE_CUDA)
...
 endif()
 
+if(PROTEUS_ENABLE_HIP)
+SET(lang HIP)
+else()
+SET(lang CUDA)
+endif()
+
+
+
```
ggeorgakoudis, line 224: Why this?

ggeorgakoudis: Makes sense, let's move this definition at top-level and remove re-setting it within individual functions

PR #292, opened 2025-09-25
PJ-DSL: While Loop

ZwFink:
This PR adds a basic while loop to the PJ-DSL (no continue or break for now).
Because the DSL has no expression-level representation, we have to explicitly re-evaluate the condition each iteration:
```c++
    auto &Cond = F.declVar<bool>("cond");
    Cond = I < N;
    F.beginWhile(Cond);
    {
      Arg[I] = Arg[I] + I;
      I = I + 1;
      Cond = I < N;
    }
    F.endWhile();
```

An alternative is to pass a lambda to generate the expression at the loop header.


tests/frontend/cpu/while.cpp
```diff
@@ -0,0 +1,66 @@
...
+  {
+    I = 0;
+    N = 10;
+    auto &Cond = F.declVar<bool>("cond");
+    Cond = I < N;
+    F.beginWhile(Cond);
+    {
+      Arg[I] = Arg[I] + I;
+      I = I + 1;
+      Cond = I < N;
```
ggeorgakoudis, line 34: Why is this necessary? Can't we re-evaluate the condition in the header?

ggeorgakoudis (commented):
LGTM

One comment

tests/frontend/cpu/while.cpp
ggeorgakoudis: I'm leaning towards this:` F.beginWhile([&](Var &cond) { cond = I < N; });`

PR #293, opened 2025-09-25
PJ-DSL: Explicit Var Casting

ZwFink:
This adds explicit Var type-casting via `FuncBase::cast`.
While the Var's `operator=` does implicit conversions:
```c++
auto &V1 = F.defVar<float>(3.5);
auto &V2 = F.declVar<int>();

V2 = V1;
```

The explicit cast operator allows you to clearly express conversion intent and is helpful when porting benchmarks from C++, as you get code that looks more similar to the original version.
```c++
auto &V1 = F.defVar<float>(3.5);
auto &V2 = F.cast<int>(V1);
```

### DIFF
--- a/tests/frontend/cpu/convert.cpp
+++ b/tests/frontend/cpu/convert.cpp
@@ -1,8 +1,8 @@
 // clang-format off
 // RUN: rm -rf "%t.$$.proteus"
-// RUN: PROTEUS_CACHE_DIR="%t.$$.proteus" %build/cast | %FILECHECK %s --check-prefixes=CHECK,CHECK-FIRST
+// RUN: PROTEUS_CACHE_DIR="%t.$$.proteus" %build/convert | %FILECHECK %s --check-prefixes=CHECK,CHECK-FIRST
 // Second run uses the object cache.
-// RUN: PROTEUS_CACHE_DIR="%t.$$.proteus" %build/cast | %FILECHECK %s --check-prefixes=CHECK,CHECK-SECOND
+// RUN: PROTEUS_CACHE_DIR="%t.$$.proteus" %build/convert | %FILECHECK %s --check-prefixes=CHECK,CHECK-SECOND
 // RUN: rm -rf "%t.$$.proteus"
 // clang-format on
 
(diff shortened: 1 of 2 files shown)


tests/frontend/cpu/cast.cpp
```diff
@@ -0,0 +1,90 @@
...
+// CHECK: RD0 = 2.5
+// CHECK-NEXT: RI0 = 3
+// CHECK-NEXT: RF0 = -7
+// CHECK-NEXT: RD1 = 3
+// CHECK-NEXT: RI1 = -7
+// CHECK-NEXT: RF1 = 2.5
+// CHECK-NEXT: RI2 = 3
+// CHECK: HashValue {{[0-9]+}} NumExecs 1 NumHits 0
+// CHECK-FIRST: JitStorageCache hits 0 total 1
+// CHECK-SECOND: JitStorageCache hits 1 total 1
```
ggeorgakoudis, line 90: It's hard to parse whether this is the expected result because the naming of the variables is very different across code. Can you come up with some sane naming to make it easy to parse the result by eyeballing?

include/proteus/Frontend/Func.hpp
```diff
@@ -237,6 +238,19 @@ class FuncBase {
     Function *F = getFunction();
     F->setName(Name);
   }
+
+  // Cast the given Var's value to type T and return a new Var holding
+  // the converted value.
+  template <typename T>
+  std::enable_if_t<std::is_arithmetic_v<T>, Var &> cast(Var &V) {
```
ggeorgakoudis, line 245: `cast` != convert. Maybe we want a more descriptive name for this operation. Why not just convert?

ggeorgakoudis (commented):
It's a little annoying that this clashes with LLVM's `cast`. Also, `cast` has overloaded semantics, in our case it means arithmetic convert. Maybe just call it `convert`? Happy to discuss

include/proteus/Frontend/Func.hpp
ZwFink: I have changed it to `convert`. I wonder if `convertTo` is even better? It delivers high clarity:
```c++
 auto &I2 = F.convertTo<int>(FfromI);
```

ggeorgakoudis (approved):
LGTM

PR #296, opened 2025-09-30
Allow BlocksPerEU to be passed as a JSON parameter

koparasy:


include/proteus/Config.hpp
```diff
@@ -229,6 +238,15 @@ class CodeGenerationConfig {
...
 
+  int blocksPerExecUnit(int MaxThreads) const {
+    // NOTE: We only return the tuned value when the current LBMaxThreads is
+    // equal to the tuned one. otherwise we return 0. Not doing so, may result
+    // in violating constraints in cases in which LBMaxThreads > TunedMaxThreads
+    if (TunedMaxThreads != MaxThreads)
+      return 0;
+    return BlocksPerExecUnit;
+  }
+
```
ggeorgakoudis, line 249: I don't understand this.

ggeorgakoudis (commented):
I don't understand this PR very well. I'll contact you offline too.

include/proteus/CompilationTask.hpp
```diff
@@ -31,6 +31,7 @@ class CompilationTask {
   CodegenOption CGOption;
   bool DumpIR;
   bool RelinkGlobalsByCopy;
+  int BlocksPerExecUnit;
```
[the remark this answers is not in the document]
koparasy: This is the second argument of launch bounds. Launch bounds when translated to llvm are mapped to different vendor concepts. Mneme and Proteus directly map to that concept. For cuda that is minBlocksPerSM for HIP that is max waves per eu. The variable itself maps to what the vendors needs

include/proteus/Config.hpp
koparasy: This is technical debt stemming from not being able to coherently compute the static hash. If Mneme tunes kernel X that's used 256 threads and then at Proteus runtime X executes with 512 threads I don't want to use this parameter.

PR #299, opened 2025-10-01
Refactor: Prefer 'if constexpr' Over Preprocessor Macros

### LINKED ISSUE #298: Avoid maybe_unused in favor of if contexpr
I'll accept but I think we have code smell here and a bad pattern with `[[maybe_unused]]`, in other parts of the code too.

We should change cmake to `add_compile_definitions(PROTEUS_ENABLE_CUDA=1)` (or set to 0, do the same for HIP) and replace checks with:

```cpp
if constexpr (PROTEUS_ENABLE_CUDA || PROTEUS_ENABLE_HIP) {
...
} else {
...
}
```

That will eliminate the ugly maybe_unused qualifiers and the preprocessor-based conditional compilation.

Please open a new PR or start an issue.

_Originally posted by @ggeorgakoudis in https://github.com/Olympus-HPC/proteus/pull/291#discussion_r2393288789_

ZwFink:
I will wait until we have CI working on an NVIDIA machine to request review.

Closes #298 

This PR replaces some uses of `#ifdef PROTEUS_ENABLE_HIP/CUDA` with `if constexpr`.
I do not apply this change in "top-level" entries outside any function/class definition.
Some instances yield compiler errors if changed; I leave them unchanged.

### DIFF
--- a/include/proteus/JitFrontend.hpp
+++ b/include/proteus/JitFrontend.hpp
@@ -74,8 +74,8 @@ private:
       if (M.isCompiled())
         PROTEUS_FATAL_ERROR("setLaunchBounds must be called before compile()");
 
-// We keep this as preprocessor 'if' because
-// setLaunchBoundsForKernel is not defined in HOST builds.
+      // We keep this as preprocessor 'if' because
+      // setLaunchBoundsForKernel is not defined in HOST builds.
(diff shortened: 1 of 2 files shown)


ZwFink:
Similar weird issues to the getAddress from the other day (even though I force push). I'm going to close this and open another

PR #309, opened 2025-10-10
PJ-DSL: Get Var Addresses

ZwFink:
Adds `getAddress` to `ScalarVar`, creating a new `PointerVar` that refers to the address.
Allows to update values within functions (need for Fey-Kac benchmark).

### DIFF
--- a/src/lib/Frontend/Var.cpp
+++ b/src/lib/Frontend/Var.cpp
@@ -834,7 +834,15 @@ Var &PointerVar::index(const Var &I) {
 }
 
 Var &PointerVar::getAddress() {
-  PROTEUS_FATAL_ERROR("PointerVar does not support getAddress");
+  // Alloca->getType() is T*.
+  auto *BasePtrTy = cast<PointerType>(Alloca->getType());
+  unsigned AddrSpace = BasePtrTy->getAddressSpace();
+  // ElemPtrTy represents the type T**.
+  Type *ElemPtrTy = PointerType::get(BasePtrTy, AddrSpace);
+  auto &ResultVar = Fn.declVarInternal("addr.", ElemPtrTy, BasePtrTy);
+  ResultVar.storePointer(Alloca);
+  return ResultVar;
+
 }
 
 ArrayVar::ArrayVar(Value *BasePointer, FuncBase &Fn, ArrayType *ArrayTy)


include/proteus/Frontend/Func.hpp
```diff
@@ -433,6 +433,35 @@ Var<std::common_type_t<T, U>> binOp(const Var<T> &L, const Var<U> &R, IntOp IOp,
   return ResultVar;
 }
 
+template <typename T>
+Var<T *> Var<T, std::enable_if_t<std::is_arithmetic_v<T>>>::getAddress() const {
```
ggeorgakoudis: Note for future todo: we probably want `std::is_scalar_v`

include/proteus/Frontend/Func.hpp
```diff
@@ -836,23 +865,58 @@ Var<T, std::enable_if_t<std::is_array_v<T>>>::operator[](
...
+    IdxValue = IRB.CreateSExtOrTrunc(IdxValue, IRB.getInt64Ty());
+
+  auto *PointerElemTy = getValueType();
+  auto *AddressBase = loadPointer();
+  unsigned AddrSpaceAddr =
+      cast<PointerType>(AddressBase->getType())->getAddressSpace();
+
+  using RetElemT = std::remove_pointer_t<T>;
+  auto &Ctx = IRB.getContext();
+  if constexpr (std::is_pointer_v<RetElemT>) {
```
ggeorgakoudis: When does this happen? Wouldn't the specialization `std::is_pointer_v` have taken over?

ggeorgakoudis (commented):
Some questions

include/proteus/Frontend/Func.hpp
```diff
@@ -836,23 +865,58 @@ Var<T, std::enable_if_t<std::is_array_v<T>>>::operator[](
...
+    auto *GEPAddr = IRB.CreateInBoundsGEP(PointerElemTy, AddressBase, IdxValue);
+    Type *SlotTy = PointerType::get(PointerElemTy, AddrSpaceAddr);
+    auto *PtrSlot = Fn.emitAlloca(SlotTy, "elem.ptr");
+    IRB.CreateStore(GEPAddr, PtrSlot);
+
+    auto Storage =
+        std::make_unique<PointerStorage>(PtrSlot, IRB, PointerElemTy);
+    return Var<RetElemT>(std::move(Storage), Fn);
```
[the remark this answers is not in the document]
ZwFink: The intention here is to match the semantics of C++. Here, we have to check whether we are going from `T**` -> `T*` (or some other instance of that) or from `T*` -> `T`

PR #311, opened 2025-10-14
PJ-DSL: Var Type Safety

ZwFink:
Var type erasure is now replaced with type safety.

Users explicitly define Vars with corresponding C++ types:
```c++
Var<int> I = F.declVar<int>();
Var<int*> Arg0 = F.getArg<0>();
Var<int[]> Arr = F.defVar<int[]>(N);
```

Type safety is the main advantage of this system:
```c++
Var<int> I = F.declVar<int>();
Var<int*> Arg0 = F.getArg<0>();

// Compiler error!
auto C = I + Arg0;

// This works: dereference the pointer.
auto C = I + *Arg0
```

Type errors lead to ahead-of-time compiler errors, like real C++ programs.

### DIFF
--- a/include/proteus/Frontend/Func.hpp
+++ b/include/proteus/Frontend/Func.hpp
@@ -87,7 +87,6 @@ public:
 
   IRBuilderBase &getIRBuilder();
 
-
   template <typename T> Var<T> declVar(StringRef Name = "var") {
     static_assert(!std::is_array_v<T>, "Expected non-array type");
 
@@ -850,7 +849,7 @@ Var<T, std::enable_if_t<std::is_pointer_v<T>>>::operator+(
 
   auto *OffsetVal = Offset.loadValue();
   auto *IdxVal = convert<OffsetT, int64_t>(IRB, OffsetVal);
-  
+
   auto *BasePtr = loadPointer();
   auto *ElemTy = getValueType();
 
@@ -875,7 +874,7 @@ Var<T, std::enable_if_t<std::is_pointer_v<T>>>::operator+(
   auto &IRB = Fn.getIRBuilder();
   auto *IntTy = IRB.getInt64Ty();
   Value *IdxVal = ConstantInt::get(IntTy, Offset);
-  
+
   auto *BasePtr = loadPointer();
   auto *ElemTy = getValueType();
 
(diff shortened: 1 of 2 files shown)


include/proteus/Frontend/Func.hpp
```diff
@@ -90,120 +87,91 @@ class FuncBase {
 
   IRBuilderBase &getIRBuilder();
 
-  // Create a variable. If PointerElemType is non-null, a PointerVar is created
-  // with an alloca-backed pointer slot; otherwise a ScalarVar is created.
-  Var &declVarInternal(StringRef Name, Type *Ty,
-                       Type *PointerElemType = nullptr);
-
-  template <typename T> Var &declVar(StringRef Name = "var") {
+  template <typename T> Var<T> declVarInternal(StringRef Name = "var") {
```
ggeorgakoudis: Can we drop `declVarInternal` and just use `declVar`?

include/proteus/Frontend/Func.hpp
```diff
@@ -90,120 +87,91 @@ class FuncBase {
...
+    return Res;
   }
 
   template <typename T>
-  Var &defRuntimeConst(T Val, StringRef Name = "run.const.var") {
-    Function *F = getFunction();
-    Var &VarRef = declVarInternal(Name, TypeMap<T>::get(F->getContext()));
-    VarRef = Val;
-    return VarRef;
+  Var<T> defRuntimeConst(T Val, StringRef Name = "run.const.var") {
```
ggeorgakoudis: Open issue for future todo: create a `ConstVar` or `ImmutableVar` instead of the loosely check `defRuntimeConst` derived class with deleted assignment methods for static checking.

include/proteus/Frontend/Func.hpp
```diff
@@ -275,15 +276,979 @@ template <typename RetT, typename... ArgT> class Func final : public FuncBase {
...
+Var<T, std::enable_if_t<std::is_arithmetic_v<T>>>::operator+(
+    const U &ConstValue) const {
+  static_assert(std::is_arithmetic_v<U>,
+                "Can only add arithmetic types to Var");
+  Var<U> Tmp = Fn.defVar<U>(ConstValue, "tmp.");
+  return (*this) + Tmp;
+}
+
+template <typename T>
+template <typename U, typename>
```
ggeorgakoudis: I find the unnamed typename not very readable. Why not do SFINAE on the return type:
```cpp
std::enable_if_t<std::is_arithmetic_v<U>, Var<std::common_type_t<T, U>>> 
```
(same change to the in-class definition)

Applies also to other arithmetic operators.

src/lib/Frontend/VarStorage.cpp
ggeorgakoudis: I'm worried that the `loadValue()`, `storeValue()` interface with `AccessKind` is error prone, but I don't have a good alternative.

Some thoughts:
1. Eliminate `AccessKind`
2. `loadValue()`, `storeValue()` function as if `AccessKind == Direct`
3. Introduce `loadValueFromPointee()`, `storeValueToPointee()` to function as if `AccessKind == Resolved, implemented only for `PointerStorage`

Or, we remove all `Access == Resolved` semantics and we delegate to lowering to explicitly emit `load`, `store` instructions from a `PointerStorage` depending on the operation context. Let's discuss

ggeorgakoudis (approved):
LGTM

PR #316, opened 2025-10-28
Allow proteus pass to be applied on non-annotated codes

koparasy:
This PR enhances the Proteus LLVM pass by enabling it to proactively annotate all GPU kernels when the relevant compile-time option is enabled. If the option is disabled, the previous behavior remains unchanged.  

## Motivation
- Proteus can now JIT kernels even when users have not manually inserted annotations, applying custom pipelines in a transparent way.  
- By exposing annotated entries as a **generic LLVM portability layer**, Proteus provides an API to access kernel LLVM IR at runtime.  
- **Mneme** leverages this capability to obtain kernel IR directly from Proteus during execution, eliminating the need to duplicate annotation logic.  

## Benefits
- Reduces code duplication in Mneme by ~1500 LOC.  
- Improves robustness of Mneme’s record/replay pipeline by relying on Proteus’ robust pass infrastructure.  
- Keeps Proteus independent, while allowing downstream tools (like Mneme) to build on its IR access layer.

### DIFF
--- a/cmake/ProteusFunctions.cmake
+++ b/cmake/ProteusFunctions.cmake
@@ -1,7 +1,20 @@
 function(add_proteus target)
-    target_compile_options(${target} PRIVATE
-        "-fpass-plugin=\$<TARGET_FILE:ProteusPass>"
-    )
+    set(options FORCE_JIT_ANNOTATE_ALL)
+    cmake_parse_arguments(PROTEUS_PASS "${options}" "" "" ${ARGN})
+
+    if (PROTEUS_PASS_FORCE_JIT_ANNOTATE_ALL)
+        target_compile_options(${target} PRIVATE
+            # -fplugin loading is needed to register the plugin command line
+            # option.
+            "-fplugin=$<TARGET_FILE:ProteusPass>"
+            "-fpass-plugin=$<TARGET_FILE:ProteusPass>"
+            "SHELL:-Xclang -mllvm -Xclang -force-proteus-jit-annotate-all"
+        )
+    else()
+        target_compile_options(${target} PRIVATE
+            "-fpass-plugin=\$<TARGET_FILE:ProteusPass>"
+        )
+    endif()
 
     target_link_options(${target} PRIVATE
         "SHELL:\$<\$<LINK_LANGUAGE:HIP>:-Xoffload-linker --load-pass-plugin=\$<TARGET_FILE:ProteusPass>>")
--- a/src/pass/AnnotationHandler.cpp
+++ b/src/pass/AnnotationHandler.cpp
@@ -267,7 +267,39 @@ static bool isLambdaFunction(const Function &F) {
 AnnotationHandler::AnnotationHandler(Module &M) : M(M), Types(M) {}
 
 void AnnotationHandler::parseAnnotations(
-    MapVector<Function *, JitFunctionInfo> &JitFunctionInfoMap) {
+    MapVector<Function *, JitFunctionInfo> &JitFunctionInfoMap,
+    const DenseMap<Value *, GlobalVariable *> &StubToKernelMap,
+    bool ForceJitAnnotateAll) {
+  // Forcing annotations overrides any user annotations and makes all kernels
+  // jit annotated without any runtime constant arguments.
+  if (ForceJitAnnotateAll) {
+    SmallVector<Constant *> ForcedJitAnnotations;
+
+    // Remove any previous user annotations.
+    removeJitGlobalAnnotations();
+
+    // Create jit annotations for all kernel functions.
+    for (auto &F : M.getFunctionList()) {
+      if (F.isDeclaration())
+        continue;
+      if (!(isDeviceKernel(&F) || StubToKernelMap.contains(&F)))
+        continue;
+      ForcedJitAnnotations.push_back(createJitAnnotation(&F, {}));
+    }
+
+    // Append the new forced annotations.
+    appendToGlobalAnnotations(ForcedJitAnnotations);
+
+    // Parse to create the function map.
+    auto *GlobalAnnotations = M.getNamedGlobal("llvm.global.annotations");
+    parseJitGlobalAnnotations(GlobalAnnotations, JitFunctionInfoMap);
+
+    // Discard after parsing.
+    removeJitGlobalAnnotations();
+
+    return;
+  }
+
   // First parse any proteus::jit_arg or proteus::jit_array annotations and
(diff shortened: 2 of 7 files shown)


koparasy:
@ggeorgakoudis would you be willing to accept something like this? This is a middle ground in respect to #315

ggeorgakoudis:
If the greatest code re-use comes from sharing the LLVM pass, then this makes much more sense to me. I'll review carefully.

koparasy:
Code reduction comes from both pass and writing code that accesses the LLVM IR at runtime.

src/pass/ProteusPass.cpp
```diff
@@ -88,14 +88,28 @@ using namespace proteus;
...
 
   bool run(Module &M, bool IsLTO) {
+    // NOTE: We need to collect early our Host Stubs, to allow aggressive
+    // annotations to be applied on all host stubs
     AnnotationHandler AnnotHandler{M};
+    if (hasDeviceLaunchKernelCalls(M)) {
+      getKernelHostStubs(M);
+    }
+
+    if (!IsLTO && AggressiveProteusAnnotation)
```
ggeorgakoudis: Why doesn't it apply to LTO compilation?

cmake/ProteusFunctions.cmake
```diff
@@ -1,7 +1,23 @@
...
+    cmake_parse_arguments(PROTEUS_PASS "${options}" "" "" ${ARGN})
+
+
+    if (PROTEUS_PASS_ENABLE_AGGRESSIVE)
+	# NOTE: LLVM seems to also support fplugin-arg-name-arg option. Documentation is limited and was added recently.
+	#Will re-investigate how do use that one
+	target_compile_options(${target} PRIVATE
+	  "-fplugin=\$<TARGET_FILE:ProteusPass>"
+          "-fpass-plugin=\$<TARGET_FILE:ProteusPass>"
+          "SHELL:-Xclang -mllvm -Xclang -enable-aggressive-proteus-annotations")
```
ggeorgakoudis: Rename `enable-aggressive-proteus-annotations` -> `enable-proteus-jit-annotate-all`

src/pass/AnnotationHandler.cpp
```diff
@@ -1274,4 +1288,122 @@ void AnnotationHandler::removeJitGlobalAnnotations() {
   AnnotationsGV->setSection("llvm.metadata");
 }
 
+Constant *AnnotationHandler::getOrCreateString(Module &M, StringRef S,
+                                               PointerType *Ty) {
+  LLVMContext &Ctx = M.getContext();
+  auto *StrConst = ConstantDataArray::getString(Ctx, S, true);
```
ggeorgakoudis: `getString(Ctx, S, /*AddNull=*/ true)`

src/pass/AnnotationHandler.cpp
```diff
@@ -1274,4 +1288,122 @@ void AnnotationHandler::removeJitGlobalAnnotations() {
   AnnotationsGV->setSection("llvm.metadata");
 }
 
+Constant *AnnotationHandler::getOrCreateString(Module &M, StringRef S,
```
ggeorgakoudis: Remove and re-use `createJitAnnotation()` function

src/pass/AnnotationHandler.cpp
```diff
@@ -1274,4 +1288,122 @@ void AnnotationHandler::removeJitGlobalAnnotations() {
...
+          if (Arr->isString() && Arr->getAsString() == Want)
+            return true;
+        }
+      }
+    }
+  }
+  return false;
+}
+
+GlobalVariable *AnnotationHandler::getOrCreateAnnoGV(Module &M) {
```
ggeorgakoudis: Re-use `appendToGlobalAnnotations()`?

src/pass/ProteusPass.cpp
```diff
@@ -88,14 +88,28 @@ using namespace proteus;
 //-----------------------------------------------------------------------------
 // ProteusPass implementation
 //-----------------------------------------------------------------------------
+
 namespace {
+static cl::opt<bool> AggressiveProteusAnnotation(
```
ggeorgakoudis: `AggressiveProteusAnnotation` -> `EnableProteusJitAnnotateAll`

src/pass/AnnotationHandler.cpp
```diff
@@ -266,6 +266,20 @@ static bool isLambdaFunction(const Function &F) {
...
 
+void AnnotationHandler::populateAnnotations(
+    DenseMap<Value *, GlobalVariable *> &StubToKernelMap) {
+
+  for (Function &F : M) {
+    if (F.isDeclaration())
+      continue;
+    if (!(isDeviceKernel(&F) || StubToKernelMap.contains(&F)))
+      continue;
+    if (hasAnnotateTag(F, "jit"))
```
ggeorgakoudis: Do you want to keep pre-existing "jit" annotations that may have pre-existing RC parameters?

src/pass/AnnotationHandler.cpp
ggeorgakoudis: Add `SmallPtrSet<Function *, 32> JitAllAnnotations;`

src/pass/ProteusPass.cpp
koparasy: Cause before LTO, we will have already annotated our functions. No need to re-annotate.

src/pass/AnnotationHandler.cpp
```diff
@@ -266,6 +266,20 @@ static bool isLambdaFunction(const Function &F) {
...
 AnnotationHandler::AnnotationHandler(Module &M) : M(M), Types(M) {}
 
+void AnnotationHandler::populateAnnotations(
+    DenseMap<Value *, GlobalVariable *> &StubToKernelMap) {
+
+  for (Function &F : M) {
+    if (F.isDeclaration())
+      continue;
+    if (!(isDeviceKernel(&F) || StubToKernelMap.contains(&F)))
+      continue;
```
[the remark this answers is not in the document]
koparasy: Cause AFAIK we apply the attribute only on GPU kernels (on both sides, device host) and we leave the rest.

src/pass/ProteusPass.cpp
ggeorgakoudis: What should happen when LTO links a module that hasn't been compiled through the Proteus pass? Should those be extracted?

ggeorgakoudis (approved):
LGTM

Some edits

PR #320, opened 2025-11-05
Support globals with same name on different TU without RDC

koparasy:
This is a test that showcases we are not handling well global variables. This is a corner case (maybe?) in which multiple translation units have global variables with the same name. Since proteus is not associating the fatbinary Handle with the global it overrides an existing entry with the latest one?


ggeorgakoudis (commented):
LGTM

Copilot comments are good to follow up.
Fix linter comments in tests.

PR #324, opened 2025-11-20
PJ-DSL For Loop: No explicit bounds type

### LINKED ISSUE #318: Enable forLoop bound type deduction
We should not need to specify `forLoop<int>(...)`. It doesn't work now, but the compiler should be able to deduce the index variable type from the constructor of `LoopBoundInfo`.

ZwFink:
The `forLoop<int>` syntax was needed so the compiler knew to construct `LoopBounds<int>`. But `initializer_list<Var<T>>` can deduce `T=int` because of initializer list rules.

Fixes #318


ggeorgakoudis (approved):
LGTM

PR #326, opened 2025-11-24
Use Return Range Metadata/Attributes Instead of `llvm.assume`

### LINKED ISSUE #307: Replace `llvm.assume` with `range` metadata
Existing codegen uses `range` metadata for similar optimization on block/grid dimensions, which we implement with assumes.

We want to investigate if `range` is better supported for optimization and change our implementation.

ZwFink:
Fixes #307 .
Return range attributes are supported in LLVM >= 19,
so for other versions we use range metadata, matching the driver's behavior
(the driver adds HW-restricted bounds, we tighten them and expand the set where they're used).


ggeorgakoudis (approved):
LGTM

Pending tests, merge

PR #337, opened 2025-12-03
PJ-DSL: Loop Unrolling

ZwFink:
Adds `.unroll(optional int Count)` to the `forLoopBuilder` to attach 'unroll' metadata to the associated loop, to match `pragma unroll`.
Unroll is not applicable to loop nests, as it's unclear what semantics you should support: apply unrolling to each loop in the nest? That seems to violate intuition from C++. Unroll only the outer loop? Then `LoopNest.tile` and `LoopNest.unroll` have different semantics.
I went with an object-oriented approach, contrasting `tile`. This is a cleaner implementation that's a better model for future loop transformations. I think refactoring `tile` is suitable for a new PR, not for this one.

I have tested the GEMM implementation on Matrix, and the DSL version now matches the annotation and CPP versions.

For now, tiling + unroll must NOT be used together (we would need to unroll all new loops? Only the outer? The inner? Unclear what you would choose to do. Non-trivial implementation). This is enforced by runtime checks.

### DIFF
--- a/src/lib/Frontend/LoopUnroller.cpp
+++ b/src/lib/Frontend/LoopUnroller.cpp
@@ -20,16 +20,15 @@ bool LoopUnroller::isEnabled() const { return Enabled; }
 void LoopUnroller::attachMetadata(llvm::BasicBlock *LatchBB) const {
   if (!Enabled)
     return;
-
-  auto *BackEdgeBr = llvm::dyn_cast<llvm::BranchInst>(LatchBB->getTerminator());
-  if (!BackEdgeBr)
-    PROTEUS_FATAL_ERROR("Expected back-edge branch in latch block");
-
   using namespace llvm;
+  auto *BackEdgeBr = dyn_cast<BranchInst>(LatchBB->getTerminator());
+
   LLVMContext &Ctx = BackEdgeBr->getContext();
 
+  // From LanguageRef: "For legacy reasons, the first item
+  // of a loop metadata node must be a reference to itself."
   SmallVector<Metadata *, 4> LoopMDOperands;
-  LoopMDOperands.push_back(nullptr); // Self-reference placeholder.
+  LoopMDOperands.push_back(nullptr);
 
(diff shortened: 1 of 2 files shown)


src/lib/Frontend/LoopUnroller.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+bool LoopUnroller::isEnabled() const { return Enabled; }
+
+void LoopUnroller::attachMetadata(llvm::BasicBlock *LatchBB) const {
+  if (!Enabled)
+    return;
+
+  auto *BackEdgeBr = llvm::dyn_cast<llvm::BranchInst>(LatchBB->getTerminator());
+  if (!BackEdgeBr)
+    PROTEUS_FATAL_ERROR("Expected back-edge branch in latch block");
```
ggeorgakoudis: This seems a superfluous check. Every basic block should have a terminator?

src/lib/Frontend/LoopUnroller.cpp
```diff
@@ -0,0 +1,51 @@
...
+  if (Count.has_value()) {
+    Metadata *CountOperands[] = {MDString::get(Ctx, "llvm.loop.unroll.count"),
+                                 ConstantAsMetadata::get(ConstantInt::get(
+                                     Type::getInt32Ty(Ctx), *Count))};
+    MDNode *UnrollCountMD = MDNode::get(Ctx, CountOperands);
+    LoopMDOperands.push_back(UnrollCountMD);
+  }
+
+  MDNode *LoopMD = MDNode::getDistinct(Ctx, LoopMDOperands);
+  LoopMD->replaceOperandWith(0, LoopMD); // Loop metadata self-reference.
```
ggeorgakoudis: Comments about. Also, make it a full sentence. Hard to understand what's the meaning

ggeorgakoudis (commented):
LGTM

Few comments

PR #356, opened 2025-12-10
PJ-DSL: Const Semantics

ZwFink:
- Adds const semantics enforcement to PJ-DSL's `Var<T>` type, making `Var<const T>` truly immutable at compile time
  - Changes `defRuntimeConst<T>()` to return `Var<const T>` instead of `Var<T>`, ensuring runtime constants cannot be accidentally
  modified
  - Refactors forLoop() API from initializer list to explicit parameters, because initializer list cannot handle mixed const/mutable types (we seem to keep having to go back and forth on this interface).

### DIFF
--- a/include/proteus/Frontend/Func.hpp
+++ b/include/proteus/Frontend/Func.hpp
@@ -5,6 +5,7 @@
 
 #include <llvm/IR/IRBuilder.h>
 #include <llvm/IR/Module.h>
+#include <type_traits>
 
 #include "proteus/AddressSpace.hpp"
 #include "proteus/Error.h"
@@ -137,10 +138,11 @@ public:
   }
 
   template <typename T, typename U>
-  Var<T> defVar(const Var<U> &Var, StringRef Name = "var") {
-    auto Res = declVar<T>(Name);
-    Res = Var;
-    return Res;
+  Var<T> defVar(const Var<U> &Val, StringRef Name = "var") {
+    using RawT = std::remove_const_t<T>;
+    Var<RawT> Res = declVar<RawT>(Name);
+    Res = Val;
+    return Var<T>(Res);
   }
 
   template <typename T>
--- a/tests/frontend/cpu/operators.cpp
+++ b/tests/frontend/cpu/operators.cpp
@@ -50,8 +50,11 @@ int main() {
     { Arg13[0] = 1.0; }
     F.endIf();
 
-    auto ConstVal = F.defVar<const int>(42, "const_val");
+    auto ConstVal = F.defVar<const int>(Cmp, "const_val");
     Arg12[0] = Arg12[0] + ConstVal;
+
+    auto ConstValScalar = F.defVar<const int>(1, "const_val_scalar");
+    Arg12[0] = Arg12[0] + ConstValScalar;
 
     auto NotCond = !(Cmp <= 5.0);
     Arg15[0] = NotCond;
@@ -105,7 +108,7 @@ int main() {
 // CHECK-NEXT: R9 = 2.5
 // CHECK-NEXT: R10 = 2
 // CHECK-NEXT: R11 = 1
-// CHECK-NEXT: R12 = 49
+// CHECK-NEXT: R12 = 13
 // CHECK-NEXT: R13 = 1
 // CHECK-NEXT: R14 = -2
 // CHECK-NEXT: R15 = 0
--- a/tests/frontend/gpu/operators.cpp
+++ b/tests/frontend/gpu/operators.cpp
@@ -69,8 +69,11 @@ int main() {
     { Arg11[0] = 1.0; }
     F.endIf();
 
-    auto ConstVal = F.defVar<const int>(42, "const_val");
+    auto ConstVal = F.defVar<const int>(Cmp, "const_val");
     Arg10[0] = Arg10[0] + ConstVal;
+
+    auto ConstValScalar = F.defVar<const int>(1, "const_val_scalar");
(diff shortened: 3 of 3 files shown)


include/proteus/Frontend/Func.hpp
```diff
@@ -207,12 +213,15 @@ class FuncBase {
...
-    auto It = Bounds.begin();
-    LoopBoundInfo<T> BoundsInfo{It[0], It[1], It[2], It[3]};
-    return ForLoopBuilder<T, BodyLambda>(BoundsInfo, *this,
-                                         std::forward<BodyLambda>(Body));
+  template <typename IterT, typename InitT, typename UpperT, typename IncT,
+            typename BodyLambda = EmptyLambda>
+  auto forLoop(Var<IterT> &Iter, const Var<InitT> &Init,
+               const Var<UpperT> &Upper, const Var<IncT> &Inc,
+               BodyLambda &&Body = {}) {
+    using RawT = std::remove_const_t<IterT>;
```
ggeorgakoudis: I certainly like this interface more. Do we have to create `RawT`? Shouldn't we error out instead if the provided iterator is specified `const`?

include/proteus/Frontend/Func.hpp
```diff
@@ -506,26 +516,32 @@ Var<bool> cmpOp(const Var<T> &L, const Var<U> &R, IntOp IOp, FPOp FOp) {
 }
 
 template <typename T>
-template <typename U>
+template <typename U, typename>
```
ggeorgakoudis: what is the unnamed typename for?

include/proteus/Frontend/Func.hpp
```diff
@@ -564,8 +580,10 @@ template <typename T>
 template <typename U>
 Var<T, std::enable_if_t<std::is_arithmetic_v<T>>> &
 Var<T, std::enable_if_t<std::is_arithmetic_v<T>>>::operator=(const Var<U> &V) {
+  static_assert(is_mutable_v<T>, "Cannot assign to Var<const T>");
   auto &IRB = Fn.getIRBuilder();
-  auto *Converted = convert<U, T>(IRB, V.loadValue());
+  auto *Converted = convert<std::remove_const_t<U>, std::remove_const_t<T>>(
```
ggeorgakoudis: `std::remove_const_t` as convenience? See previous comment.

include/proteus/Frontend/Var.hpp
```diff
@@ -103,9 +103,11 @@ struct Var<T, std::enable_if_t<std::is_arithmetic_v<T>>>
   Var(std::unique_ptr<VarStorage> Storage, FuncBase &Fn)
       : VarStorageOwner<VarStorage>(std::move(Storage), Fn) {}
 
-  // // Conversion constructor
-  // // TODO: Add an is_convertible check.
-  template <typename U> Var(const Var<U> &V);
+  // Conversion constructor from Var<U> where U can convert to T.
+  template <typename U, typename = std::enable_if_t<
+                            std::is_convertible_v<U, std::remove_const_t<T>> &&
```
ggeorgakoudis: Should it be also `std::remove_const_t<U>`?

include/proteus/Frontend/Func.hpp
```diff
@@ -138,8 +144,8 @@ class FuncBase {
   }
 
   template <typename T>
-  Var<T> defRuntimeConst(const T &Val, StringRef Name = "run.const.var") {
-    return defVar<T>(Val, Name);
+  Var<const T> defRuntimeConst(const T &Val, StringRef Name = "run.const.var") {
+    return Var<const T>(defVar<T>(Val, Name));
```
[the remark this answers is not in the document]
ZwFink: I vote to keep `defRuntimeConst` to keep the explicit API for creating runtime constants rather than rely on `defVar`'s double-duty abilities.

include/proteus/Frontend/Func.hpp
```diff
@@ -506,26 +516,32 @@ Var<bool> cmpOp(const Var<T> &L, const Var<U> &R, IntOp IOp, FPOp FOp) {
...
+  auto &IRB = Fn.getIRBuilder();
+  using RawU = std::remove_const_t<U>;
+  auto *Converted = convert<RawU, RawT>(IRB, V.loadValue());
+  storeValue(Converted);
 }
 
 template <typename T>
 Var<T, std::enable_if_t<std::is_arithmetic_v<T>>> &
 Var<T, std::enable_if_t<std::is_arithmetic_v<T>>>::operator=(const Var &V) {
+  static_assert(is_mutable_v<T>, "Cannot assign to Var<const T>");
```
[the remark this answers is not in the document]
ZwFink: If we could, I would make all type errors come from `static_assert`. I think it's way more ergonomic for users to get these bespoke error messages.
I think it's worthwhile to accept this proliferation of `static_asserts` in these cases.

tests/frontend/cpu/operators.cpp
```diff
@@ -53,6 +50,9 @@ int main() {
     { Arg13[0] = 1.0; }
     F.endIf();
 
+    auto ConstVal = F.defVar<const int>(42, "const_val");
```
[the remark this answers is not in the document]
ZwFink: Hmm, yes. Maybe it's better to change this test to `defVar<const int>` something that's a DSL variable that we want to make unmutable.

ggeorgakoudis (approved):
LGTM

Left the discussion on the runtime const function naming open for you to close.

PR #368, opened 2025-12-22
PJ-DSL: Reference Semantics

### LINKED ISSUE #313: PJ-DSL: Reference Semantics
Dereferencing a `Var<T*>` returns a `Var<T>`, a scalar whose storage wraps the pointee.
This is misleading, as `Var<T>` generally uses value semantics.
Instead, dereferencing a `Var<T*>` should return a `Var<T&>`.

ZwFink:
Closes #313 .
Adds DSL interface support for `Var<T&>`.
Now, array and pointer `Var` de-reference operators return a `Var<T&>`, where they previously returned a `Var<T>`.
`Var<T&>.getAddress()` returns a `Var<T*>` to match C++ semantics.


include/proteus/Frontend/TypeTraits.hpp
```diff
@@ -0,0 +1,36 @@
+#ifndef PROTEUS_FRONTEND_TYPETRAITS_HPP
+#define PROTEUS_FRONTEND_TYPETRAITS_HPP
+
+#include <type_traits>
+
+namespace proteus {
+
+// Type alias to remove cv-qualifiers and references.
+template <typename T>
+using clean_t = std::remove_cv_t<std::remove_reference_t<T>>;
```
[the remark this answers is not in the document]
ggeorgakoudis: Actually C++20 includes it: https://en.cppreference.com/w/cpp/types/remove_cvref.html

include/proteus/Frontend/TypeTraits.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+// NOLINTBEGIN(readability-identifier-naming)
+
+// True if T (after removing references) is an arithmetic type.
+template <typename T>
+inline constexpr bool is_arithmetic_unref_v =
+    std::is_arithmetic_v<std::remove_reference_t<T>>;
```
ggeorgakoudis: Is `std::remove_reference_t` enough or do we also want to drop `cv` qualifiers? I assume we use the trait only for rvalue computation?

include/proteus/Frontend/TypeTraits.hpp
ggeorgakoudis: More general comment and probably a different PR, but does it make sense to encode runtime constants by a `const` qualified `Var` type instead of the special `defRuntimeConst` interface?

include/proteus/Frontend/Var.hpp
```diff
@@ -192,21 +194,22 @@ struct Var<T, std::enable_if_t<std::is_array_v<T>>>
...
 
-// Specialization for pointer types
+// Specialization for pointer types (including references to pointers)
 template <typename T>
-struct Var<T, std::enable_if_t<std::is_pointer_v<T>>>
+struct Var<T, std::enable_if_t<is_pointer_unref_v<T>>>
     : public VarStorageOwner<PointerStorage> {
   using ValueType = T;
-  using ElemType = std::remove_pointer_t<T>;
+  using ElemType = std::remove_pointer_t<std::remove_reference_t<T>>;
```
ggeorgakoudis: Are we re-defining `clean_t`?

tests/frontend/cpu/operators.cpp
```diff
@@ -90,6 +112,10 @@ int main() {
   std::cout << "R13 = " << R13 << "\n";
   std::cout << "R14 = " << R14 << "\n";
   std::cout << "R15 = " << R15 << "\n";
+  std::cout << "R16 = " << R16 << "\n";
+  std::cout << "R17 = " << R17 << "\n";
+  std::cout << "R18 = " << R18 << "\n";
+  std::cout << "R19 = " << R19 << "\n";
```
ggeorgakoudis: I think the test would be more useful if we print through the original storage, to make sure the reference modifies that

include/proteus/Frontend/Func.hpp
```diff
@@ -419,10 +423,12 @@ class FuncBase {
...
   template <typename U, typename T>
-  std::enable_if_t<std::is_convertible_v<T, U>, Var<U>>
+  std::enable_if_t<std::is_convertible_v<std::remove_reference_t<T>,
+                                         std::remove_reference_t<U>>,
+                   Var<clean_t<U>>>
   convert(const Var<T> &V) {
-    Var<U> Res = declVar<U>("convert.");
-    Value *Converted = convert<T, U>(V.loadValue());
+    Var<clean_t<U>> Res = declVar<clean_t<U>>("convert.");
+    Value *Converted = convert<clean_t<T>, clean_t<U>>(V.loadValue());
```
ggeorgakoudis: Instead of passing `clean_t` types to convert, can't we just apply `clean_t` in the `convert` definition to avoid forgetting doing the "cleaning" at each callsite?

ggeorgakoudis (commented):
LGTM

Some questions and comments

include/proteus/Frontend/Func.hpp
```diff
@@ -419,10 +423,12 @@ class FuncBase {
...
   // the converted value.
   template <typename U, typename T>
-  std::enable_if_t<std::is_convertible_v<T, U>, Var<U>>
+  std::enable_if_t<std::is_convertible_v<std::remove_reference_t<T>,
+                                         std::remove_reference_t<U>>,
+                   Var<clean_t<U>>>
   convert(const Var<T> &V) {
-    Var<U> Res = declVar<U>("convert.");
-    Value *Converted = convert<T, U>(V.loadValue());
+    Var<clean_t<U>> Res = declVar<clean_t<U>>("convert.");
```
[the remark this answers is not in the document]
ZwFink: No. We should drop only ref, cv should remain

tests/frontend/cpu/operators.cpp
ggeorgakoudis: nvm, brain freeze, all is well

tests/frontend/gpu/operators.cpp
```diff
@@ -125,6 +152,10 @@ int main() {
   std::cout << "R11 = " << *R11 << "\n";
   std::cout << "R12 = " << *R12 << "\n";
   std::cout << "R13 = " << *R13 << "\n";
+  std::cout << "R14 = " << *R14 << "\n";
+  std::cout << "R15 = " << *R15 << "\n";
+  std::cout << "R16 = " << *R16 << "\n";
+  std::cout << "R17 = " << *R17 << "\n";
```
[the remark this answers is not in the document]
ggeorgakoudis: nvm

Issue #365, opened 2025-12-16
Re-think symbol naming for JIT-generated functions


ggeorgakoudis:
Why is this done? To avoid name collisions with multiple specializations of a function?

_Originally posted by @ZwFink in https://github.com/Olympus-HPC/proteus/pull/360#discussion_r2624253395_

PR #372, opened 2026-01-07
Demangleable JIT Kernel/Function Names

ZwFink:
Change JIT kernel name suffix from `$jit$<hash>$` to `.jit.<hash>` to enable proper demangling by standard tools.

The previous suffix format `$jit$<hash>$` produced names that failed to demangle, making it challenging to use performance analysis tools such as `rocprof`.

Example existing kernel that doesn't demangle:
```bash
  $ echo '_Z6kernelPiPf$jit$12345$' | c++filt
  _Z6kernelPiPf$jit$12345$
```

Now the names demangle correctly:
```bash
  $ echo '_Z6kernelPiPf.jit_12345' | c++filt
  kernel(int*, float*) [clone .jit.12345]

  $ echo '_ZN2ns6kernelEPiPf.jit_12345' | llvm-cxxfilt
  ns::kernel(int*, float*) (.jit.12345)
```

This uses the existing "." suffix for [function cloning](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=40831), which seems appropriate to me as JIT introduces a new means to create function clones.
This works in Rocprof, where the output looks like: `void mfem::HipKernel1D<mfem::Vector::operator*=(double)::{lambda(int)#1}>(int, mfem::Vector::operator*=(double)::{lambda(int)#1}) [clone .jit.12347350548553885757]` (Rocprof uses `abi::__cxa_demangle`).

This fix is only compatible with host + HIP because PTX identifiers cannot contain '.'
This is a step toward #365. Ideally, the demangled information should include specialized names and values.


ZwFink:
Creates invalid PTX kernel names: https://docs.nvidia.com/cuda/parallel-thread-execution/#identifiers

johnbowen42 (approved):
love it

PR #374, opened 2026-01-08
Fix LLVM static initialization with shared proteus

koparasy:
This PR is a draft reproducer for a runtime failure observed when Mneme links against Proteus as a shared library linking LLVM statically and also directly links with LLVM libraries.

## Context

An application uses LLVM APIs (e.g., Mneme), the application links Proteus as a shared library, and the process ends up with more than one LLVM instance.

This violates LLVM’s one-LLVM-per-process assumption and results in runtime failures such as:

```bash
: CommandLine Error: Option 'amdgpu-dump-hsa-metadata' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options
Aborted (core dumped)
```

What this PR contains

A minimal reproducer test that:
* uses LLVM in the application,
* links Proteus as a shared library,
* triggers the double-LLVM failure.

The test is intentionally simple and exists solely to document and reproduce the issue seen with Mneme.


ggeorgakoudis:
I'm also fine to take this over this one, by @koparasy's permission

ggeorgakoudis (approved):

PR #375, opened 2026-01-10
Debloat host JIT modules

ggeorgakoudis:
Host JIT modules can be bloated because our internalization and expected DCE when creating them is sub-optimal, letting functions to escape as externals, e.g., functions in COMDAT groups such as constructors.

This PR improves internalization removing unnecessary COMDAT and uses our own transitive closure cloning, instead of whole module cloning, to avoid bloating.


tomstitt (approved):
Thanks @ggeorgakoudis ! Our big object files are now 1000x smaller 😎

PR #383, opened 2026-01-15
Build shared library alongside static when enabled

ggeorgakoudis:
- Allows clients (mneme) to link with the shared or static proteus library based on usage
- Update add_proteus helper with LINK_SHARED option
- Update tests


koparasy (approved):
LGTM, you should tag maybe the PR that I have active about this?

PR #384, opened 2026-01-18
Link only needed from LLVM and privatize

### LINKED ISSUE #359: [Refactor] Avoid leaking LLVM headers to consumers

ggeorgakoudis:
Link only necessary LLVM libraries using `llvm_map_components_to_libnames`.

We should not expose consumers to LLVM library, definitions, or headers.

Closes #359


ggeorgakoudis:
Rebase to main ☝️

PR #386, opened 2026-01-20
PJ-DSL: Define/Declare Multiple Vars

ZwFink:
Adds `declVars` and `defVars` to the PJ-DSL, similar to `defRuntimeConsts`.
It's a handy shorthand for the beginning of kernels when you define/declare a bunch of variables.
Array variables are not supported.

This PR also adds support for when you want to copy a `Var` and give the new copy a name, e.g.,:
```c++
Var<double> A = F.defVar(0.5, "A");
Var<double> A_Copy = F.defVar(A, "A_Copy");
```
The current way to copy a Var with `operator=` does not support the above.

### DIFF
--- a/include/proteus/Frontend/Func.h
+++ b/include/proteus/Frontend/Func.h
@@ -334,10 +334,17 @@ public:
     return Var<T>(Res);
   }
 
+  template <typename U>
+  Var<U> defVar(const Var<U> &Val, const std::string &Name = "var") {
+    Var<U> Res = declVar<U>(Name);
+    Res = Val;
+    return Res;
+  }
+
   template <
       typename T, typename NameT,
       typename = std::enable_if_t<std::is_convertible_v<NameT, std::string>>>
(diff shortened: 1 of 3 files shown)


ggeorgakoudis (commented):
Can we add naming to this interface? It'd be nice to name variable for readability of the output LLVM IR.

Also, let's make a new test instead of hijacking `for`.

ggeorgakoudis:
> > Can we add naming to this interface? It'd be nice to name variable for readability of the output LLVM IR.
> > We can do this for `declVars` easily.
> > For `defVar`, I think the easiest option is `std::pair` to create a named value:
> 
> ```c++
> F.defVars(std::pair{10, "I"}, 20, std::pair{30, "K"});
> ```
> 
> This allows us to mix and match named with unnamed `Var`s. I think `std::pair` is better than a bespoke class, something like `NamedVarInitializer<T>(T Val, std::string Name = "...")`.

How will we typecheck `std::pair`?

PR #390, opened 2026-01-29
Use pdebug for matrix ci

ggeorgakoudis:


ZwFink (approved):
LDNTM! (Looks Desperately Needed To Me)

PR #393, opened 2026-02-03
bugfix:  Fix jit_variables associating with incorrect lambdas bug

johnbowen42:
Previously, the lambda value specialization assumed that a lambda with `jit_variable` in its capture list would be immediately called--which is typically the case for portability layers like RAJA. However, if a user declared multiple lambdas on the stack, and launched them in any order, the first launched lambda would register all calls to `jit_variable` as its own.  This did not create correctness issues, but did result in Proteus failing to specialize for user-provided runtime constants for all but the first declared lambda.  `lambda_spec_test` shows this

This PR is a bugfix, but it significantly changes how our lambda runtime value specialization works.  Roughly, `ProteusPass` now uses use-def analysis to associate a `jit_variable` call --> `class.anon` type emitted by clang.  We change `register_lambda` to force an allocation of the same `class.anon` type.  Using this pattern, we can associate `jit_variable`-->demangled, clang emitted lambda name.  Then, we can intercept execution of these lambdas at runtime, using the appropriate runtime values.

### DIFF
--- a/tests/cpu/lambda_spec_test.cpp
+++ b/tests/cpu/lambda_spec_test.cpp
@@ -13,7 +13,7 @@
 template <typename LambdaType> class Abstraction {
 public:
   LambdaType Lambda;
-  Abstraction(const LambdaType &Lam) : Lambda(Lam) {};
+  Abstraction(const LambdaType &Lam) : Lambda(Lam){};
   auto operator()() { return Lambda(); }
 };
 
@@ -32,14 +32,20 @@ int main() {
   int One = 1;
   int Two = 2;
 
-  auto ZeroLambda = [=, C = proteus::jit_variable(Zero)]()
-                        __attribute__((annotate("jit"))) { printInt(C); };
+  auto ZeroLambda = [ =, C = proteus::jit_variable(Zero) ]()
+      __attribute__((annotate("jit"))) {
+    printInt(C);
+  };
 
-  auto OneLambda = [=, C = proteus::jit_variable(One)]()
-                       __attribute__((annotate("jit"))) { printInt(C); };
+  auto OneLambda =
+      [ =, C = proteus::jit_variable(One) ]() __attribute__((annotate("jit"))) {
+    printInt(C);
+  };
 
-  auto TwoLambda = [=, C = proteus::jit_variable(Two)]()
-                       __attribute__((annotate("jit"))) { printInt(C); };
+  auto TwoLambda =
+      [ =, C = proteus::jit_variable(Two) ]() __attribute__((annotate("jit"))) {
+    printInt(C);
+  };
 
   run(ZeroLambda);
   run(OneLambda);
--- a/tests/gpu/lambda_spec_test.cpp
+++ b/tests/gpu/lambda_spec_test.cpp
@@ -30,14 +30,20 @@ int main() {
   int One = 1;
   int Two = 2;
 
-  auto ZeroLambda = [=, C = proteus::jit_variable(Zero)]()
-                        __attribute__((annotate("jit"))) { printInt(C); };
+  auto ZeroLambda = [ =, C = proteus::jit_variable(Zero) ]()
+      __attribute__((annotate("jit"))) {
+    printInt(C);
+  };
 
-  auto OneLambda = [=, C = proteus::jit_variable(One)]()
-                       __attribute__((annotate("jit"))) { printInt(C); };
+  auto OneLambda =
+      [ =, C = proteus::jit_variable(One) ]() __attribute__((annotate("jit"))) {
+    printInt(C);
+  };
 
-  auto TwoLambda = [=, C = proteus::jit_variable(Two)]()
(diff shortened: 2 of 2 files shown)


src/include/proteus/impl/LambdaRegistry.h
```diff
@@ -52,22 +52,38 @@ class LambdaRegistry {
     return It;
   }
 
-  void pushJitVariable(RuntimeConstant &RC) {
-    PendingJitVariables.emplace_back(RC);
+  void dump() {
+    llvm::outs() << "DUMPING TMP REGISTRY \n";
```
ggeorgakoudis: Use the `Logger` interface

src/pass/ProteusPass.cpp
```diff
@@ -1238,6 +1240,132 @@ class ProteusPassImpl {
...
+          }
+        }
+      }
+    }
+    // Erase the calls, downstream optimization passes should erase the alloca
+    // as well
+    for (CallBase *CB : ToErase) {
+      CB->eraseFromParent();
+    }
+  }
```
ggeorgakoudis: Simplify, `__jit_take_address` is an extern "C" which doesn't return a value. We should be able to do:
```
auto *JitTakeAddress = M.getFuction("__jit_take_address");
SmallVector<CallBase *> ToDelete;
for(User *Usr : JitTakeAddress.users()) {
  CallBase *CB = dyn_cast<Callbase>(Usr);
  assert(CB && "Expected only call users");
  ToDelete.push_back(CB);
}
...
```

src/runtime/JitEngineHost.cpp
```diff
@@ -187,11 +188,11 @@ void getLambdaJitValues(StringRef FnName,
   SmallVector<StringRef> LambdaCalleeInfo;
   PROTEUS_DBG(Logger::logs("proteus")
               << " Trying F " << demangle(FnName.str()) << "\n ");
-  auto OptionalMapIt = LambdaRegistry::instance().matchJitVariableMap(FnName);
+  auto OptionalMapIt = LR.matchJitVariableMap(FnName);
   if (!OptionalMapIt)
     return;
 
-  LambdaJitValuesVec = OptionalMapIt.value()->getSecond();
+  LambdaJitValuesVec = OptionalMapIt.value()->second;
```
ggeorgakoudis: unnecessary change

src/pass/ProteusPass.cpp
johnbowen42: https://github.com/Olympus-HPC/proteus/issues/418

src/pass/ProteusPass.cpp
```diff
@@ -1374,7 +1505,13 @@ class ProteusPassImpl {
...
         IRBuilder<> Builder(CB);
         auto *LambdaNameGlobal = Builder.CreateGlobalString(LambdaType);
-        CB->setArgOperand(1, LambdaNameGlobal);
+        bool HasSRETArg = false;
+        for (uint32_t I = 0; I < CB->getNumOperands(); ++I) {
+          HasSRETArg =
+              HasSRETArg || CB->paramHasAttr(I, llvm::Attribute::StructRet);
+        }
+        int LambdaNameIndex = HasSRETArg ? 2 : 1;
+        CB->setArgOperand(LambdaNameIndex, LambdaNameGlobal);
```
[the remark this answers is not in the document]
ggeorgakoudis: Can avoid this code now that the `register_lambda` is `void`?

tests/cpu/lambda_spec_test.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <cstdio>
+
+#include <proteus/JitInterface.h>
+
+template<typename LambdaType>
+class Abstraction {
+  public:
+  LambdaType Lambda;
+  Abstraction(const LambdaType& Lam) : Lambda(Lam) {};
+  // Abstraction(const LambdaType& Lam) : Lambda(Lam) {};
```
ggeorgakoudis: Remove comment

src/pass/ProteusPass.cpp
```diff
@@ -1239,6 +1241,115 @@ class ProteusPassImpl {
...
+            JitVarFunctions.push_back(CB);
+          }
+        }
+      }
+    }
+
+    for (CallBase *CB : JitVarFunctions) {
+      // traverse use-def from each callsite, the CB value will be used by the
+      // allocated lambda class.
+      std::queue<Value *> WorkList;
```
ggeorgakoudis: `std::queue` -> `SmallVector`, we already heavily use this data structure

src/include/proteus/impl/LambdaRegistry.h
```diff
@@ -52,22 +52,40 @@ class LambdaRegistry {
     return It;
   }
 
-  void pushJitVariable(RuntimeConstant &RC) {
-    PendingJitVariables.emplace_back(RC);
+  void dump() {
```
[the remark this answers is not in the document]
johnbowen42: yes

PR #400, opened 2026-02-10
Completely Centralized MPI Cache

ZwFink:
MPI Cache where a communication thread on rank 0 is both the reader and writer.

- Refactor monolithic `MPISharedStorageCache` into a shared base class (`MPIStorageCache`) with common MPI helpers (`MPIHelpers`)
- Add `MPILocalLookupCache`: rank 0 serves disk lookups, other ranks look up locally then fall back to rank 0
- Add `MPIRemoteLookupCache`: fully centralized cache where all ranks (including rank 0) make synchronous MPI requests to the rank-0 communication thread for both store and lookup

### DIFF
--- a/src/runtime/Caching/MPIStorageCache.cpp
+++ b/src/runtime/Caching/MPIStorageCache.cpp
@@ -82,8 +82,6 @@ void MPIStorageCache::finalize() {
 
   CommThread.join();
 
-  // MPI_Barrier(Comm);
-
   CommHandle.free();
   Finalized = true;
 }


src/include/proteus/impl/Caching/MPIHelpers.h
```diff
@@ -0,0 +1,101 @@
...
+#include <mpi.h>
+
+#include <memory>
+#include <thread>
+#include <vector>
+
+namespace proteus {
+
+enum class MPITag : int {
+  Store = 0,
```
ggeorgakoudis: nit: maybe start with a non-zero value to help debugging and detect missed init errors

src/runtime/Caching/MPIHelpers.cpp
```diff
@@ -0,0 +1,164 @@
...
+}
+
+MPICommHandle::~MPICommHandle() {
+  if (Comm == MPI_COMM_NULL)
+    return;
+
+  int MPIFinalized = 0;
+  MPI_Finalized(&MPIFinalized);
+  if (!MPIFinalized)
+    MPI_Comm_free(&Comm);
```
ggeorgakoudis: Error out if MPI is not finalized, it should never happen

src/runtime/Caching/MPIRemoteLookupCache.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+std::unique_ptr<CompiledLibrary>
+MPIRemoteLookupCache::lookup(const HashT &HashValue) {
+  TIMESCOPE("MPIRemoteLookupCache::lookup");
+  Accesses++;
+
+  // Rank 0 reads directly from disk to avoid sending a LookupRequest to its
+  // own comm thread, which would race with MPI_Probe(ANY_TAG).
+  auto Result = CommHandle.getRank() == 0 ? lookupFromDisk(HashValue)
+                                          : lookupRemote(HashValue);
```
ggeorgakoudis: Do we need this special treatment of rank 0? There shouldn't be a race given the comm thread is independent.

Also break symmetry with the write operation

ggeorgakoudis (commented):
LGTM.

Some comments.

src/runtime/Caching/MPIStorageCache.cpp
```diff
@@ -0,0 +1,244 @@
...
+  MPI_Comm Comm = CommHandle.get();
+  int Err = MPI_Isend(Pending->Buffer.data(),
+                      static_cast<int>(Pending->Buffer.size()), MPI_BYTE,
+                      /*dest=*/0, static_cast<int>(MPITag::Store), Comm,
+                      &Pending->Request);
+  if (Err != MPI_SUCCESS) {
+    reportFatalError("MPI_Isend failed with error code " + std::to_string(Err));
+  }
+
+  PendingSends.push_back(std::move(Pending));
```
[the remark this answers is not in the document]
ZwFink: We want to store the whole PendingSend object so we are ensured the buffer lives long enough

src/runtime/Caching/MPIRemoteLookupCache.cpp
ggeorgakoudis: OK, I see. Rank 0 is special.

src/runtime/Caching/MPIRemoteLookupCache.cpp
```diff
@@ -0,0 +1,238 @@
...
+  return Result;
+}
+
+std::unique_ptr<CompiledLibrary>
+MPIRemoteLookupCache::lookupRemote(const HashT &HashValue) {
+  MPI_Comm Comm = CommHandle.get();
+
+  auto ReqBuf = packLookupRequest(HashValue);
+  MPI_Send(ReqBuf.data(), static_cast<int>(ReqBuf.size()), MPI_BYTE, 0,
+           static_cast<int>(MPITag::LookupRequest), Comm);
```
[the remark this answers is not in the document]
ggeorgakoudis: it is required, `enum class`...

ggeorgakoudis:
Rebase on latest main ☝️

PR #403, opened 2026-02-11
Kernel JIT Tracing

ZwFink:
Adds an option `PROTEUS_KERNEL_TRACE` to produce an end-of-run summary of kernels that were JIT compiled. This is helpful for applications like MFEM where you want to quickly see readable JIT statistics without having to do rocprof -> analysis.

End of run summary looks like this:
```
[proteus][JitEngineDevice] === Kernel Trace (rank 0) ===
[proteus][JitEngineDevice]   kernel()  rank=0 specializations=5  launches=5
[proteus][JitEngineDevice] === End Kernel Trace ===
```

In the future, it would be nice to have a printout of all kernels that were JIT-compiled and all that weren't. This requires changing the pass, however so I leave it to the future.


src/include/proteus/impl/Caching/MemoryCache.h
```diff
@@ -66,11 +65,39 @@ template <typename Function_t> class MemoryCache {
...
+      return;
+
+    if (CacheMap.empty())
+      return;
+
+    struct KernelStats {
+      size_t Specializations = 0;
+      uint64_t TotalLaunches = 0;
+    };
+    std::map<std::string, KernelStats> Grouped;
```
ggeorgakoudis: `unordered_map`?

ggeorgakoudis (approved):
LGTM

nit: stencil.cpp trips linter checking

PR #407, opened 2026-02-18
Ensure instrumentation of all kernel function registrations

koparasy:
There are cases where a translation defines a kernel but does call it. We still need to instrument the registration.


src/pass/ProteusPass.cpp
```diff
@@ -136,8 +136,9 @@ class ProteusPassImpl {
     findJitVariables(M);
     registerLambdaFunctions(M);
 
+    instrumentRegisterFunction(M);
```
ggeorgakoudis: nit: Move it together with other `instrument*` calls, e.g., after `instrumentRegisterVar(M)`

PR #464, opened 2026-04-02
Automatically mark read-only lambda captures as constant

davidbeckingsale:


tests/cpu/lambda_field_pointer_escape.cpp
```diff
@@ -0,0 +1,38 @@
...
+  proteus::register_lambda(lambda);
+  lambda();
+
+  std::cout << "x[0] = " << X[0] << "\n";
+  std::cout << "x[1] = " << X[1] << "\n";
+
+  return 0;
+}
+
+// CHECK-DAG: [LambdaSpec][Auto] Replacing slot {{[0-9]+}} with double 3.14
```
johnbowen42, line 35: I see B is only read in the lambda--does that mean it should be constant in IR?  If so, it would be nice if this message said that explicitly in the test

src/pass/AutoReadOnlyCapturesAnalysis.cpp
```diff
@@ -0,0 +1,529 @@
...
+constexpr char AutoReadOnlyCapturesMetadataName[] =
+    "proteus.auto_readonly_captures";
+
+struct SlotState {
+  int32_t ByteOffset;
+  RuntimeConstantType RCType = RuntimeConstantType::NONE;
+  bool IsReadOnly = true;
+};
+
+enum class PointerUseEffect {
```
johnbowen42, line 39: could we potentially replace this with the existing `ModRefResult` enum?

src/pass/AutoReadOnlyCapturesAnalysis.cpp
```diff
@@ -0,0 +1,529 @@
...
+
+  if (auto *SI = dyn_cast<StoreInst>(U)) {
+    if (SI->getPointerOperand() == Ptr)
+      return PointerUseEffect::WriteOrEscape;
+    if (SI->getValueOperand() == Ptr)
+      return PointerUseEffect::WriteOrEscape;
+    return PointerUseEffect::Ignore;
+  }
+
+  if (isa<CallBase>(U))
```
johnbowen42, line 152: Could add an easy check here for if the Ptr is passed as a `const` value to the CB

src/pass/AutoReadOnlyCapturesAnalysis.cpp
```diff
@@ -0,0 +1,529 @@
...
+          DEBUG(Logger::logs("proteus-pass")
+                << "[AutoReadOnly] Slot " << SlotIndex
+                << " disqualified: written by store: " << *SI << "\n");
+          markSlotNonReadOnly(Slots, SlotIndex, ByteOffset);
+          break;
+        }
+        continue;
+      }
+
+      PointerUseEffect Effect = classifyPointerUse(U, V);
```
johnbowen42, line 317: Can't we just call `getModRefInfo` here on the current `Instruction`

src/pass/AutoReadOnlyCapturesAnalysis.cpp
```diff
@@ -0,0 +1,529 @@
...
+      if (Effect == PointerUseEffect::Ignore ||
```
johnbowen42, line 318: similarly here you would check the `ModRefInfo` enum struct

src/pass/AutoReadOnlyCapturesAnalysis.cpp
```diff
@@ -0,0 +1,529 @@
...
+        if (!LocalOffset) {
+          DEBUG(Logger::logs("proteus-pass")
+                << "[AutoReadOnly] Skipping GEP with non-constant offset: "
+                << *GEP << "\n");
+          continue;
+        }
+
+        auto TopLevelSlot = getTopLevelSlotIndex(ClosureArg, V, GEP);
+        int32_t ByteOffset = *LocalOffset;
+        analyzePointerUsersForSlot(F, ClosureArg, GEP,
```
johnbowen42, line 462: We might even be able to simplify this DFS analysis by collecting all the "slots", or any GEP/load into the `ClosureArg`, and simply do
```
DenseSet<Slot> isModified;
for (Slot : Slots) {
  for (auto &BB : F) {
      for (auto &I : BB) {
        if (getModRefInfo(Slot, I) != ModRefInfo::Ref)
          continue;
        isModified.insert(Slot);
       }
    }
}
  
```
then we mark as const an slot that isn't contained within `isModified`.  I might be missing something here but I think it  might simplify this analysis and reuse existing LLVM pointer analysis tools, which would both be good.  It's possible LLVM is too conservative with `getModRefInfo` for our purposes here, which would be interesting

src/pass/ProteusPass.cpp
```diff
@@ -795,7 +799,16 @@ class ProteusPassImpl {
 
     ArrayType *ArgPtrsTy = ArrayType::get(Types.PtrTy, StubFn->arg_size());
     Value *ArgPtrs = nullptr;
-    if (NumRuntimeConstants > 0) {
+
+    // Check if this is a lambda function (contains ::operator() in demangled
+    // name)
+    std::string DemangledName = llvm::demangle(StubFn->getName().str());
```
johnbowen42, line 805: This will need to change a little bit with the new lambda handling.  Firstly, we have both LambdaFunctorWrapper::operator() methods, as well as underlying lambda::operator() methods.  This check will pass for both, as well as any `operator()` named function, which we don't want.  I would simply change this to check if the metadata contains the string `proteus.wrapper_call`, which indicates we are handling a registered functor

johnbowen42 (commented):
I think we should see if using `getModRefInfo` gets us good checks on if a "slot"`, or GEP into a closure pointer, is read/written.  If it gives us reasonable results, I think we should use that method.  This would greatly simplify the analysis here and simplify/eliminate the def-use analysis

PR #467, opened 2026-04-14
feature: redesign Proteus's lambda JIT compilation interface and pipeline

johnbowen42:
This feature restructures Proteus's lambda support so that a device module containing multiple calls of a lambda with different runtime constants can be correctly JIT compiled.  This feature makes Proteus's support of JIT compiled lambdas closer to supporting all possible C++ lambda syntax, instead of the narrow syntax often used by parallel programming abstraction layers.  Concretely

- Add a functor wrapper to Proteus lambdas so that lambda callsites can be transparently tracked
- Add LambdaPass which runs at the start of the pipeline and (1) handles jit_variable callsite analysis on less-simplified IR (before InstCombine can cause indirection at callsite (2) instruments lambda runtime constant registration for host lambda callsites and (3) generates manifests of lambda constants registered to device lambda callsite.  Also LambdaPass removes global annotations on lambda functor operator callsites and attaches them as metadata nodes
- Change ProteusPass so that (1) it consumes the LambdaPass manifests and then (2) uses them to build lambda registration functions, which are executed right before kernel launch
- Change the runtime to look for lambda callsites in a device module containing metadata nodes, use that metadata as a key for lookup, and specialize each callsite based on the result lookup.


cmake/ProteusFunctions.cmake
ggeorgakoudis, line 1: I think installing and using a separate pass is quirky. We already register ProteusPass in multiple places (early-simplifications and LTO), so we can register it also at pipeline-start and just do the lambda analysis

include/proteus/Frontend/TypeTraits.h
ggeorgakoudis, line 1: This header is for serving the CPP/DSL frontends. Why do we have these changes here?

include/proteus/JitInterface.h
```diff
@@ -47,6 +57,7 @@ jit_array(T V, [[maybe_unused]] size_t NumElts,
           typename std::remove_pointer<T>::type Velem = 0) noexcept;
 #endif
 
+
```
ggeorgakoudis: Stray newline

scripts/setup-rocm.sh
```diff
@@ -22,6 +22,7 @@ cmake .. \
 -DPROTEUS_ENABLE_HIP=on \
 -DCMAKE_INSTALL_PREFIX=../install-${HOSTN}-rocm-${ROCM_VERSION} \
 -DCMAKE_EXPORT_COMPILE_COMMANDS=on \
+-DENABLE_TESTS=ON \
```
ggeorgakoudis: Stray change

src/include/proteus/impl/Cloning.h
```diff
@@ -555,6 +557,10 @@ inline std::unique_ptr<Module> cloneKernelFromModules(
   auto KernelModule = Cloner.cloneClosure(*EntryM, EntryF->getContext(),
                                           Reachable, ShouldCloneDefinition);
 
+  // Ensure `proteus.*` annotations are retained even though
+  // `llvm.global.annotations` is not part of the reachability closure.
+  rebuildGlobalAnnotationsByPrefixFromModules(Mods, *KernelModule, "proteus.");
+
```
ggeorgakoudis: I don't think llvm.global.annotations is the right interface to preserve this information. They have appending linkage (grows big) and it's easy for conflicts to arise with other plugins (e.g., enzyme). We drop them in ProteusPass for good reasons. We can attach a metadata node to the lambda functions and scan the IR.

src/include/proteus/impl/Cloning.h
ggeorgakoudis, line 1: I don't think we should modify this file

src/include/proteus/impl/GlobalAnnotations.h
ggeorgakoudis, line 1: I think llvm.global.annotations is the wrong way to preserve this information. I propose metadata on function

src/include/proteus/impl/CompilationTask.h
```diff
@@ -25,7 +25,7 @@ class CompilationTask {
   dim3 BlockDim;
   dim3 GridDim;
   SmallVector<RuntimeConstant> RCVec;
-  SmallVector<std::pair<std::string, StringRef>> LambdaCalleeInfo;
+  SmallVector<std::pair<Function*, uint64_t>> LambdaCalleeInfo;
```
ggeorgakoudis: This confuses me. `Function` is the DSL representation of a function. Mixing this for lambda identification for annotations is not desirable.

ggeorgakoudis (commented):
This is great and will solve our lambda issues.

I made a first pass and left comments on things we need to restructure.

include/proteus/Frontend/TypeTraits.h
johnbowen42: I had been adding some function signature type trait helpers in an earlier draft of the functor work, but it ended up being unnecessary so I'll revert these changes

src/include/proteus/impl/Cloning.h
johnbowen42: I was debating doing this but ended up not going with it because of the requisite scan of all module functions.  But we can cache this ultimately and only do it once per runtime.

src/include/proteus/impl/CompilationTask.h
johnbowen42: previously, the `LambdaCalleeInfo` contained the mangled names of all lambdas.  Now, instead of passing around those StringRefs we pass around the `Function*`.  The ID identifies which `LambdaFunctorWrapper` is associated with the `Function*`

src/include/proteus/impl/GlobalAnnotations.h
johnbowen42: gotcha.  Happy to delete this boilerplate

johnbowen42: I changed this so that the annotation gets attached to the function as metadata node

src/include/proteus/impl/Cloning.h
johnbowen42: modified to make sure function metadata is cloned

cmake/ProteusFunctions.cmake
johnbowen42: Could you expand on this?  How else would be run an early LambdaAnalysis pass and Proteuspass later?

johnbowen42:
all tests are passing for me locally on Tioga with rocm 6.4.2

cmake/ProteusFunctions.cmake
ggeorgakoudis: Register ProteusPass at pipeline-start and pass an argument when you construct it to indicate the insert point. Check how we do it for early-simplifications and LTO. Apparently state is not true-false anymore, if there are 3 possible insert points.

src/include/proteus/impl/Cloning.h
ggeorgakoudis: `CloneFunctionInto` already copies metadata

src/include/proteus/impl/TransformLambdaSpecialization.h
```diff
@@ -122,31 +131,450 @@ class TransformLambdaSpecialization {
...
+    default:
+      reportFatalError("Unsupported RuntimeConstantType in kernel-arg reader");
+    }
+
+    return RC;
+  }
+
+
+public:
+  static void transform(Module &M,
```
johnbowen42: I added a new, faster approach for specializing lambda operators that need different variants for different tuples of runtime constants.  The new file `KernelArgVisitor.h` add a visitor that analyzes pointer uses from callsite to kernel launch.  It identifies the raw offset from the base pointer passed to the kernel in bytes and then this method reads them off the kernel args.

src/pass/LambdaAnalysis.cpp
```diff
@@ -0,0 +1,967 @@
...
+  bool run(Module &M, bool) {
+    // Attach Metadata nodes corresponding to llvm.global.annotations to
+    // individual function calls.
+    addMetadataNodeWithNameFromLLVMGlobal(M, "proteus.wrapper_call");
+    addMetadataNodeWithNameFromLLVMGlobal(M, "proteus.register_call");
+    addMetadataNodeWithNameFromLLVMGlobal(M, "proteus.register_call_impl");
+    // We need collect any kernel host stubs to pass to parse annotations, used
+    // in forced annotations.
+    const auto StubToKernelMap = getKernelHostStubs(M);
+
```
ZwFink: Copy-paste artifact from ProteusPass? This is never used here. Can be removed along with getKernelHostStubs.

src/pass/LambdaAnalysis.cpp
```diff
@@ -0,0 +1,967 @@
...
+    for (auto &KV : FunctionsToAnnotate) {
+      Function *F = KV.first;
+      uint64_t Id = KV.second;
+      setRegisteredLambdaMetadata(F, Id);
+    }
+  }
+
+  DenseMap<Value *, GlobalVariable *> getKernelHostStubs(Module &M) {
+    DenseMap<Value *, GlobalVariable *> StubToKernelMap;
```
ZwFink: Remove

src/pass/LambdaAnalysis.cpp
```diff
@@ -0,0 +1,967 @@
...
+  }
+
+private:
+  ProteusTypes Types;
+
+  MapVector<Function *, JitFunctionInfo> JitFunctionInfoMap;
+  SmallPtrSet<Function *, 16> ModuleDeviceKernels;
+
+  Value *getStubGV([[maybe_unused]] Value *Operand) {
+    // NOTE: when called by isDeviceKernelHostStub, Operand may not be a global
```
ZwFink: Remove

src/pass/LambdaAnalysis.cpp
```diff
@@ -0,0 +1,967 @@
...
+             Builder.getInt32(Idx), JitVarInfo.Offset, FieldPtr,
+             Builder.getInt64(ID)});
+      }
+      // Insert the finalize call
+      Builder.CreateCall(FinalizeFn, {Builder.getInt64(ID)});
+    }
+  }
+
+  bool hasDeviceLaunchKernelCalls(Module &M) {
+    Function *LaunchKernelFn = nullptr;
```
ZwFink: Remove

src/pass/LambdaAnalysis.cpp
```diff
@@ -0,0 +1,967 @@
...
+    addMetadataNodeWithNameFromLLVMGlobal(M, "proteus.wrapper_call");
+    addMetadataNodeWithNameFromLLVMGlobal(M, "proteus.register_call");
+    addMetadataNodeWithNameFromLLVMGlobal(M, "proteus.register_call_impl");
+    // We need collect any kernel host stubs to pass to parse annotations, used
+    // in forced annotations.
+    const auto StubToKernelMap = getKernelHostStubs(M);
+
+    DenseMap<StructType *, SmallVector<LambdaJitVariableInfo, 16>>
+        LambdaStorageTypeToJitIndices;
+    DenseMap<Type *, GlobalVariable *> LambdaTypeToGlobalName;
```
ZwFink: Not used

src/pass/LambdaAnalysis.cpp
```diff
@@ -0,0 +1,967 @@
...
+      reportFatalError("Broken original module found, compilation aborted!");
+
+    return true;
+  }
+
+private:
+  ProteusTypes Types;
+
+  MapVector<Function *, JitFunctionInfo> JitFunctionInfoMap;
+  SmallPtrSet<Function *, 16> ModuleDeviceKernels;
```
ZwFink: Unused

src/pass/LambdaAnalysis.cpp
```diff
@@ -0,0 +1,967 @@
...
+
+// New PM implementation.
+struct LambdaPass : PassInfoMixin<LambdaPass> {
+  LambdaPass(bool IsLTO) : IsLTO(IsLTO) {}
+  bool IsLTO;
+
+  PreservedAnalyses run(Module &M, ModuleAnalysisManager & /*AM*/) {
+    LambdaPassImpl PPI{M};
+
+    bool Changed = PPI.run(M, IsLTO);
```
ZwFink: The `run` method just ignores LTO. Copy-paste artifact?

src/pass/LambdaAnalysis.cpp
```diff
@@ -0,0 +1,967 @@
...
+      NewArgIt->setName(OldArg.getName());
+      VMap[&OldArg] = &(*NewArgIt++);
+    }
+
+    SmallVector<ReturnInst *, 8> Returns;
+    CloneFunctionInto(NewFunc, LambdaOperator, VMap,
+                      CloneFunctionChangeType::LocalChangesOnly, Returns);
+
+    return NewFunc;
+  }
```
ZwFink, line 317:
```suggestion
  }

```
Add newline

src/pass/LambdaAnalysis.cpp
```diff
@@ -0,0 +1,967 @@
...
+    // in forced annotations.
+    const auto StubToKernelMap = getKernelHostStubs(M);
+
+    DenseMap<StructType *, SmallVector<LambdaJitVariableInfo, 16>>
+        LambdaStorageTypeToJitIndices;
+    DenseMap<Type *, GlobalVariable *> LambdaTypeToGlobalName;
+    DenseSet<StructType *> RegisteredLambdaStorageClasses;
+    DenseMap<StructType *, StructType *> FunctorTypeToLambdaTypeMap;
+    DenseMap<uint64_t, StructType *> FunctorIDToLambdaTypeMap;
+    DenseMap<uint64_t, StructType *> FunctorIDToFunctorTypeMap;
```
ZwFink: These are all declared here but used only for host compilation?

johnbowen42:
This analysis is now compatible with RAJA launch, kernel, and forall LLVM IR shapes
https://github.com/llnl/RAJA/pull/2052

src/pass/KernelArgVisitor.h
```diff
@@ -0,0 +1,643 @@
...
+      if (!Analysis || Analysis->KernelArgIndex != BaseSlot ||
+          Analysis->Offset != BaseOffset) {
+        AnalysisFailed = true;
+        AnalysisSuccess = false;
+        return;
+      }
+    }
+    KernelFunction = FirstAnalysis->KernelFunction;
+    AnalysisSuccess = true;
+    AnalysisFailed = false;
```
ZwFink, line 582: Don't you also need to set `KernelArg`, `Offset`, and `ChangedRC`?

src/pass/KernelArgVisitor.h
```diff
@@ -0,0 +1,643 @@
...
+    }
+
+    for (User *U : F->users()) {
+      auto *CB = dyn_cast<CallBase>(U);
+      if (!CB)
+        continue;
+      DEBUG(Logger::logs("proteus-pass")
+            << "Analysis crossed interprocedural boundary at "
+            << *CB->getArgOperand(ArgNum) << "\n");
+      WorkList.push_back({CB->getArgOperand(ArgNum), CB});
```
ZwFink, line 542: Does this need to be context sensitive?

src/pass/LambdaAnalysis.cpp
```diff
@@ -0,0 +1,874 @@
...
+    if (Ty->isIntegerTy(64))
+      return RuntimeConstantType::INT64;
+    if (Ty->isFloatTy())
+      return RuntimeConstantType::FLOAT;
+    if (Ty->isDoubleTy())
+      return RuntimeConstantType::DOUBLE;
+    if (Ty->isPointerTy())
+      return RuntimeConstantType::PTR;
+    return std::nullopt;
+  }
```
ZwFink, line 531: Regression: long double is dropped

src/pass/ProteusPass.cpp
ZwFink: Should the lambda analysis be run here too?

ZwFink (changes requested):
This is great work. I have mostly minor things.
Because this PR is so big and has been so long in the works, I don't want to hold it up for any of these things. 
I am concerned that we would perpetually run around in bike-shedding circles, all the while this important feature is held up. Anything you choose not to resolve should be opened as an issue for follow-up.

ZwFink (approved):
LGTM! woohoo

davidbeckingsale (approved):

PR #468, opened 2026-04-16
Support MLIR code as frontend input

ggeorgakoudis:
- Move MLIR to LLVM IR lowering from MLIRCodeBuilder in the MLIRLower component to make it re-usable
- Implement MLIRJitModule that takes MLIR source as a string
- Add elementary tests


src/runtime/Frontend/MLIRJitModule.cpp
```diff
@@ -0,0 +1,94 @@
...
+  if (Verify && verifyModule(*Lowered.Mod, &llvm::errs()))
+    reportFatalError("Broken module found, JIT compilation aborted!");
+
+  llvm::SmallVector<char, 0> Buffer;
+  llvm::raw_svector_ostream OS(Buffer);
+  llvm::WriteBitcodeToFile(*Lowered.Mod, OS);
+  ModuleHash = std::make_unique<HashT>(
+      hash(llvm::StringRef{Buffer.data(), Buffer.size()}));
+
+  if ((Library = Dispatch.lookupCompiledLibrary(*ModuleHash))) {
```
ZwFink: By that I mean: Can we hash the MLIR + other stuff and check that before lowering to LLVM?

src/runtime/Frontend/MLIRLower.cpp
```diff
@@ -0,0 +1,494 @@
...
+      NVVM::NVVMDialect,
+#endif
+#if PROTEUS_ENABLE_HIP
+      ROCDL::ROCDLDialect,
+#endif
+      math::MathDialect, memref::MemRefDialect, scf::SCFDialect>();
+}
+
+MLIRLoweringResult lowerMLIRModuleToLLVM(mlir::ModuleOp Module,
+                                         const MLIRLoweringOptions &Options) {
```
ZwFink: This is a candidate for refactor. >300 line function with multiple phases applied to distinct architectures

ZwFink (approved):
LGTM. Up to you whether it's worth addressing comments.

include/proteus/MLIRJitModule.h
```diff
@@ -0,0 +1,118 @@
...
+#include <string>
+#include <type_traits>
+#include <utility>
+
+namespace proteus {
+
+struct CompiledLibrary;
+class HashT;
+
+class MLIRJitModule {
```
ZwFink, line 18: Is this exposed in non-MLIR builds?

src/runtime/Frontend/MLIRLower.cpp
```diff
@@ -0,0 +1,494 @@
...
+#endif
+  }
+
+  Operation *OperationToSerialize = Module.getOperation();
+  if (TM == TargetModelType::CUDA || TM == TargetModelType::HIP) {
+    auto DeviceSym = Module.lookupSymbol<gpu::GPUModuleOp>("kernels");
+    if (!DeviceSym)
+      reportFatalError(Prefix.str() +
+                       ": expected gpu.module @kernels for device lowering");
+    mlir::PassManager DevicePM(Context);
```
ZwFink: This assumption of `gpu.module @kernels` is now user-facing. Is it documented?

src/runtime/Frontend/MLIRJitModule.cpp
ggeorgakoudis: I think we can. I'm mulling over that, but it seems a worthy change because it will avoid the lowering overhead

include/proteus/MLIRJitModule.h
ggeorgakoudis: It shouldn't. What is the issue?

src/runtime/Frontend/MLIRLower.cpp
ggeorgakoudis: No, it's not. It got carried over when I moved the code to `MLIRLower`. At minimum we need to document, at best I'd like not to have this restriction for user-provided code in the `MLIRJitModule`

include/proteus/MLIRJitModule.h
ggeorgakoudis: Oh yeah, we always install it.

PR #477, opened 2026-04-22
Support PROTEUS_OPT_PIPELINE across LLVM JIT paths

### LINKED ISSUE #274: [Docs] Add that PROTEUS_OPT_PIPELINE is possible only with PROTEUS_CODEGEN=serial or parallel

### LINKED ISSUE #379: Hashing should include the specialization environment to avoid stale kernels in the cache
Contributed by @koparasy 

Proteus hashes should include the specializations applied to the kernel.

For example, running

```bash
$ PROTEUS_SPECIALIZE_LAUNCH_BOUNDS=0 ./exe
$ PROTEUS_SPECIALIZE_LAUNCH_BOUNDS=1 ./exe
```

Second run will find a kernel in the cache with unspecialized launch bounds and won't recompile it to apply the specialization

ggeorgakoudis:
Make Proteus-owned LLVM optimization paths consistently use the configured optimization pipeline. This includes frontend compileOnly paths, CppJit Clang LLVM IR compilation, and HIP parallel LTO codegen.

Also update cache keys so pipeline-affecting codegen configuration invalidates the right artifacts, while keeping runtime specialization policy separate from frontend module hashes. Document the resulting optimization pipeline behavior and support matrix in the developer internals docs.

Closes #379 
Closes #274

### DIFF
--- a/docs/dev/optimization-pipeline.md
+++ b/docs/dev/optimization-pipeline.md
@@ -1,0 +1,116 @@
+# Optimization Pipeline
+
+This page documents where `PROTEUS_OPT_PIPELINE` applies inside Proteus.
+
+`PROTEUS_OPT_PIPELINE` is a textual LLVM pass pipeline. It is used when Proteus
+owns LLVM IR optimization, or when Proteus can forward the textual pipeline to
+an LLVM API such as LTO. It is not a generic compiler flag for external
+compiler paths such as NVCC, HIPRTC, or Clang shared-library compilation.
+
+## Selection Order
+
+Proteus selects the optimization pipeline from the effective
+`CodeGenerationConfig`:
+
+1. A per-kernel JSON `"Pipeline"` from `PROTEUS_TUNED_KERNELS`, when the code
+   path looks up config by kernel name.
+2. The global `PROTEUS_OPT_PIPELINE` environment variable.
+3. The default pipeline implied by `PROTEUS_OPT_LEVEL`.
+
+Per-kernel JSON `"Pipeline"` currently applies to annotated CUDA/HIP runtime
+JIT paths because they call:
+
+```cpp
+Config::get().getCGConfig(KernelName)
+```
+
+Frontend module compilation currently uses the global config:
+
+```cpp
+Config::get().getCGConfig()
+```
+
+so per-kernel JSON `"Pipeline"` does not naturally apply to DSL, MLIR, or C++
+frontend modules.
+
+## Support Matrix
+
+| Frontend / API path | Target type | Main compile path | Uses `PROTEUS_OPT_PIPELINE`? | Notes |
+| --- | --- | --- | --- | --- |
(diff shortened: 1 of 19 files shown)


docs/dev/optimization-pipeline.md
```diff
@@ -0,0 +1,116 @@
...
+```cpp
+Config::get().getCGConfig(KernelName)
+```
+
+Frontend module compilation currently uses the global config:
+
+```cpp
+Config::get().getCGConfig()
+```
+
```
ZwFink, line 32: This is an odd limitation. We should open an issue to fix it

src/runtime/Frontend/CppJitCompilerClang.cpp
```diff
@@ -125,12 +125,18 @@ class CppJitCompilerClang : public CppJitCompiler {
...
+    // runs its configured middle-end pipeline when the dispatcher compiles the
+    // returned LLVM IR.
     if (Request.TargetModel == TargetModelType::HOST) {
       ArgStorage = {PROTEUS_CLANGXX_BIN,
                     "-emit-llvm",
                     "-S",
                     "-std=c++17",
                     CppJitCompiler::FrontendOptLevelFlag,
+                    "-Xclang",
+                    "-disable-llvm-passes",
```
ZwFink, line 139: Does this solve our problem of "how do you get un-optimized, but optimizable code out of clang"?

src/include/proteus/impl/CoreLLVMHIP.h
```diff
@@ -461,7 +469,8 @@ codegenObject(Module &M, StringRef DeviceArch,
     ObjectFiles = detail::codegenSerial(M, DeviceArch);
     break;
   case CodegenOption::Parallel:
-    ObjectFiles = detail::codegenParallel(M, DeviceArch);
+    ObjectFiles = detail::codegenParallel(M, DeviceArch, /*OptLevel=*/3,
+                                          /*CodegenOptLevel=*/3, OptPipeline);
     break;
```
ZwFink, line 474: Seems in the spirit of the PR to pass through OptimizationPipelineConfig to this method and use the opt levels from there

ZwFink (approved):

docs/dev/optimization-pipeline.md
ggeorgakoudis: Yes, this path is Mneme-specific too and we definitely need a better interface. Let's discuss out-of-band.

src/runtime/Frontend/CppJitCompilerClang.cpp
ggeorgakoudis: By my micro-tests yes, but I think we want to do a test equivalent to the GEMM issue you faced before.

ggeorgakoudis: Verified on the GEMM benchmark that code is identical and no performance degradation exists.

PR #478, opened 2026-04-24
Split ProteusPass offload plugin and remove IRLinker

ggeorgakoudis:
- Remove the unused IRLinker shim and link the pass directly.
- Build a separate ProteusPassOffload plugin that links with needed LLVM libraries for HIP RDC use
- Add a stamp file for tracking ProteusPass dependency in test files to decouple from building internals for ProteusPass


src/pass/CMakeLists.txt
```diff
@@ -30,29 +6,57 @@ set(PROTEUS_PASS_SOURCES
...
+
+add_library(ProteusPassOffload SHARED
+  $<TARGET_OBJECTS:ProteusPassObjects>
+)
+
 # Allow undefined symbols in shared objects on Darwin (this is the default
 # behaviour on Linux)
 target_link_libraries(ProteusPass
-  PRIVATE IRLinker
   "$<$<PLATFORM_ID:Darwin>:-undefined dynamic_lookup>")
```
ZwFink: Should we still link ProteusPass as `PRIVATE` here?

ZwFink (approved):

PR #479, opened 2026-04-25
Add Python bindings for Proteus JIT modules

ggeorgakoudis:
- Add CMake option and build rules for pybind11-based bindings
- Expose C++ and MLIR frontends through a unified Python API
- Support host function calls and GPU kernel launches from Python
- Add host/GPU Python smoke and validation tests
- Fix RTTI handling for the runtime target and binding integration
- Enable testing for python bindings


bindings/python/PythonBindings.cpp
```diff
@@ -0,0 +1,461 @@
...
+
+  std::string Text = py::str(PathLike);
+  std::ifstream Input(Text);
+  if (!Input)
+    return Text;
+
+  std::ostringstream OS;
+  OS << Input.rdbuf();
+  return OS.str();
+}
```
ZwFink, line 47: Comment is misleading here as this will also accept the string `kernel.cpp`

bindings/python/PythonBindings.cpp
```diff
@@ -0,0 +1,461 @@
...
+  } catch (const py::error_already_set &) {
+    PyErr_Clear();
+  }
+  try {
+    // `ctypes.addressof()` covers explicit ctypes instances and buffers that do
+    // not expose array-style protocols.
+    return ctypes.attr("addressof")(Obj).cast<uintptr_t>();
+  } catch (const py::error_already_set &) {
+    PyErr_Clear();
+  }
```
ZwFink, line 118: Unfortunate how messy Python interfaces for pointers are

bindings/python/PythonBindings.cpp
```diff
@@ -0,0 +1,461 @@
...
+
+  if (py::hasattr(Obj, "data_ptr"))
+    // Match tensor libraries that surface a callable `data_ptr()` helper.
+    return Obj.attr("data_ptr")().cast<uintptr_t>();
+
+  if (py::hasattr(Obj, "__cuda_array_interface__"))
+    // Reuse device pointers from GPU array-like objects exposing the CUDA
+    // array interface protocol.
+    return ptrFromInterface(Obj, "__cuda_array_interface__");
+
```
ZwFink, line 139: Might be worth checking for DLPack also? IIUC, that is intended to replace cuda_array_interface

ZwFink (approved):
LGTM, aside from some nits. There is no way to install the bindings for use, I assume that's what #480 and a presumed GPU followup will be for?

bindings/python/PythonBindings.cpp
ggeorgakoudis: Agreed. I will elevate to an issue.

bindings/python/PythonBindings.cpp
ggeorgakoudis: Tell me about it

bindings/python/PythonBindings.cpp
ggeorgakoudis: Agreed. I will elevate to an issue

PR #484, opened 2026-04-28
Add LLVM IR JIT module and frontend tests

ggeorgakoudis:
- Add `LLVMIRJitModule` support for compiling host and device LLVM IR from source or bitcode.
- Add CPU and GPU lit coverage for source, bitcode, and cache-hit behavior.
- Add python bindings


ZwFink (approved):
LGTM!

PR #485, opened 2026-04-28
Resolve CUDA toolchain at runtime

ggeorgakoudis:
- Detect CUDA roots from runtime env vars
- Resolve libdevice and libcudart paths from the selected toolkit
- Validate runtime CUDA toolkit major version against the build toolkit
- Add docs

### DIFF
--- a/src/runtime/Frontend/CUDAToolchain.cpp
+++ b/src/runtime/Frontend/CUDAToolchain.cpp
@@ -1,0 +1,258 @@
+#include "proteus/impl/Frontend/CUDAToolchain.h"
+
+#include "proteus/Error.h"
+
+#if PROTEUS_ENABLE_CUDA
+
+#include "proteus/impl/Config.h"
+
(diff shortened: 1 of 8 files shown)


src/runtime/Frontend/CUDAToolchain.cpp
```diff
@@ -0,0 +1,258 @@
...
+  static std::optional<ResolvedCUDAToolchain> Cache;
+
+  if (Cache)
+    return *Cache;
+
+  std::lock_guard<std::mutex> Lock(CacheMutex);
+  if (!Cache)
+    Cache = resolveCUDAToolchainImpl();
+  return *Cache;
+}
```
ZwFink: Isn't this a race? First `if (Cache)` should be protected by mutex

ZwFink (changes requested):

ZwFink (approved):

PR #487, opened 2026-04-29
Resolve HIP bitcode assets from ROCm at runtime

ggeorgakoudis:
- Add a cached HIP toolchain resolver using PROTEUS_ROCM_PATH or ROCM_PATH
- Derive the ROCm amdgcn/bitcode directory at runtime
- Validate runtime HIP major/minor version against the build HIP version
- Switch DispatcherHIP to load ROCm bitcode through the runtime resolver
- Remove configure-time AMDDeviceLibs bitcode path embedding
- Add docs on HIP runtime asset detection

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,29 +82,7 @@ if(PROTEUS_ENABLE_HIP)
   message(STATUS "hiprtc Version: ${hiprtc_VERSION}")
   if(hip_VERSION VERSION_LESS "6.2.0")
     message(FATAL_ERROR "HIP found: ${hip_VERSION} is less than minimum required version 6.2.0.")
-	endif()
-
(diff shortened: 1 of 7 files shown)


ZwFink (approved):
Fair bit of duplicated functions between this and the CUDA version (`getNonEmptyEnvVar`, `canonicalizeExistingPath` and some others), but I'm not sure this matters

ggeorgakoudis:
> Fair bit of duplicated functions between this and the CUDA version (`getNonEmptyEnvVar`, `canonicalizeExistingPath` and some others), but I'm not sure this matters

Yes, I agree. I plan to refactor once all the functionality is upstream and tested for a baseline. A similar refactor is due for the XXXJitModule variants we have as they share common code.

PR #488, opened 2026-04-30
Split wheels into shim and backend packages with CUDA support

ggeorgakoudis:
- Add a pure-python shim package with default host backend dependency
- Add host and cuda 12 backend wheel manifests and loader selection
- Add linux cuda wheel support using runtime libcuda loading
- Restructure wheel ci into separate shim, host, and cuda jobs
- Move python wheel image scripts under packaging/python/image-scripts
- Update wheel tests and documentation


src/runtime/CUDADriverAPI.cpp
```diff
@@ -0,0 +1,176 @@
...
+template <typename Fn> Fn resolveCUDADriverSymbol(const char *Name) {
+  void *Handle = getCUDADriverHandle();
+  dlerror();
+  void *Symbol = dlsym(Handle, Name);
+  if (const char *Err = dlerror()) {
+    proteus::reportFatalError("Failed to resolve CUDA driver symbol " +
+                              std::string(Name) + ": " + Err);
+  }
+  return reinterpret_cast<Fn>(Symbol);
+}
```
ZwFink, line 44: You may want to use `cuGetProcAddress` first here to resolve string to correctly resolve runtime function names to their correct version suffix when applicable. https://docs.nvidia.com/cuda/archive/11.5.2/cuda-driver-api/group__CUDA__DRIVER__ENTRY__POINT.html

ZwFink (changes requested):
It looks like GPU tests aren't run for the CUDA wheel build in CI. This seems like it will be easy to miss bugs + crashes in the wheel version.

ggeorgakoudis:
> It looks like GPU tests aren't run for the CUDA wheel build in CI. This seems like it will be easy to miss bugs + crashes in the wheel version.

It's intentional at this point because setting up testing in gitlab CI with wheels will be a later PR

src/runtime/CUDADriverAPI.cpp
ggeorgakoudis: Good catch. Though I think the solution is to use the symbol from `cuda.h` to avoid messy versioning and `PFN_*` type churn, and boostrapping `cuGetProcAddress`. Check commit https://github.com/Olympus-HPC/proteus/pull/488/commits/dad8d013b39ece95b91a61ac14d0927527c9c9ba

ZwFink (approved):
LGTM

PR #489, opened 2026-05-06
Expose Kernel Metadata Independent of LLVM

ZwFink:
Access to kernel metadata is granted to Mneme via LLVM data structures, causing DSO boundary crossing and crashes. This branch exposes the metadata through an independent interface, allowing Proteus to share data with Mneme without crossing LLVM boundaries.


ggeorgakoudis (approved):
LGTM

Fix tests, we merge

PR #493, opened 2026-05-08
Load ROCm HIP libraries at runtime

ggeorgakoudis:
We avoid the dependency at build time to simplify python packaging.


ZwFink (approved):
LGTM!

PR #496, opened 2026-05-22
feature: Add interface for plugin pass

johnbowen42:
We want our lambda interface to be compatible with Enzyme.  Towards this end, this PR adds an option to run a user-supplied plugin pass as part of the optimization pipeline at JIT optimization time


johnbowen42:
@jandrej this PR is the first step in towards solving compatibility between Proteus and Enzyme

ZwFink:
Why only for serial codegen?

src/runtime/JITPassPluginRegistry.cpp
```diff
@@ -0,0 +1,95 @@
...
+    if (llvm::sys::fs::status(Config.Path, Status))
+      return Config.Path + "|" + Config.Pipeline;
+
+    llvm::SmallString<128> Storage;
+    llvm::raw_svector_ostream OS(Storage);
+    const llvm::sys::fs::UniqueID ID = Status.getUniqueID();
+    OS << Config.Path << "|" << Config.Pipeline << "|" << ID.getDevice() << ":"
+       << ID.getFile() << "|" << Status.getSize() << "|"
+       << Status.getLastModificationTime().time_since_epoch().count();
+    return std::string(OS.str());
```
ZwFink: Doesn't this mean that the MPI caches cannot share code? Because the hash is now dependent on unique rank state.

src/runtime/JITPassPluginRegistry.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <llvm/Support/FileSystem.h>
+#include <llvm/Support/raw_ostream.h>
+
+#include <algorithm>
+#include <mutex>
+
+namespace proteus {
+namespace {
+
+class JITPassPluginRegistry {
```
ZwFink, line 17: Should the registry be scoped to a JITEngine instead of being global?

src/include/proteus/impl/CoreLLVM.h
```diff
@@ -88,9 +97,57 @@ createTargetMachine(Module &M, StringRef Arch, unsigned OptLevel = 3) {
   return TM;
 }
 
-inline void runOptimizationPassPipeline(Module &M, StringRef Arch,
-                                        const std::string &PassPipeline,
-                                        unsigned CodegenOptLevel = 3) {
+inline std::string getDefaultOptimizationPipeline(char OptLevel) {
```
ZwFink, line 100: This switch also exists in runOptimizationPassPipeline. That should just call this one

src/runtime/JITPassPluginRegistry.cpp
```diff
@@ -0,0 +1,95 @@
...
+void registerJITPassPluginImpl(const std::string &PluginPath,
+                               const std::string &PassPipeline) {
+  JITPassPluginRegistry::instance().registerPlugin(PluginPath, PassPipeline);
+}
+
+void clearJITPassPluginsImpl() { JITPassPluginRegistry::instance().clear(); }
+
+std::vector<JITPassPluginConfig> getJITPassPluginConfigs() {
+  return JITPassPluginRegistry::instance().snapshot();
+}
```
ZwFink, line 97: This adds global lock to JIT kernel launches, even when no plugins are loaded. It would be helpful to add an atomic check of whether any plugins are loaded to avoid this.

ZwFink (changes requested):
Nice changes! I have a couple small issues

johnbowen42:
> Why only for serial codegen?

@ZwFink We only ever call `optimizeIR` if `CodegenOption::Serial` is set: https://github.com/Olympus-HPC/proteus/blob/ea4dfde87809b2652a4a04f928d869590f82a0e8/src/include/proteus/impl/CompilationTask.h#L63

src/runtime/JITPassPluginRegistry.cpp
johnbowen42: Interesting question.  The LambdaRegistry has the same scope and is not scoped within the JITEngine.  I think the current visibility is fine as long as it mirrors the LambdaRegistry's, but we might want to change both to be scoped within JitEngine.  I do think we want the JITPassPluginRegistry to be more "persistent" than the LambdaRegistry

src/include/proteus/impl/CoreLLVM.h
johnbowen42: I think technically the one you're talking about returns enum values, this one return strings.  Correct me if I misunderstood

ggeorgakoudis:
LGTM, but I'll delegate acceptance to @ZwFink, who has a more recent view of the codebase. I see some trivial issues with tests (linter and matrix was probably down)

Does it solve the issue with including the enzyme plugin?

ZwFink (approved):

Issue #495, opened 2026-05-19
Releasing Spack package


cmelone:
How would you feel about submitting the Proteus Spack package to the builtin repository? I just wrote one for Mneme and this would make it easier to install them both

ggeorgakoudis:
@cmelone I think it's a great idea and will help distribute Proteus more widely. Have you experience with upstreaming to Spack builtin repo? How does maintenance work?

cmelone:
It's mostly a matter of submitting the existing package to the https://github.com/spack/spack-packages repo. You can designate a list of maintainers; at least one of them must approve PRs that change the package.

Once it's in the builtin repo, there's little need to maintain a version of the package in this repo. If you'd like to go that direction, I can submit a PR for the package in #499 and we can go from there.

ggeorgakoudis:
Sounds reasonable. I'll delegate technical scrutiny to @ZwFink and @johnbowen42, who are going to be doing most of the technical work.

cmelone:
Closing the loop on this:

- the spack package was upstreamed in https://github.com/spack/spack-packages/pull/5519
- I discussed the Spack regression test strategy with a Spack maintainer and they said it's best to keep the development package in the local repository (ie this proteus repo), and upstream to spack-packages as necessary

PR #499, opened 2026-06-24
Spack packaging updates

cmelone:
Update package repo to use the api v2.2 layout, add new release, and require c/cxx from LLVM providers.

It might be worth upstreaming this package to the builtin Spack repo to make it easier for dependents like Mneme to use Proteus

also, declare `__proteus_launch_kernel` weakly so mneme can link to proteus statically

### DIFF
--- a/packaging/spack/spack_repo/proteus/packages/proteus/package.py
+++ b/packaging/spack/spack_repo/proteus/packages/proteus/package.py
@@ -5,7 +5,7 @@
 
 from spack_repo.builtin.build_systems.cuda import CudaPackage
 from spack_repo.builtin.build_systems.rocm import ROCmPackage
-
+from spack_repo.builtin.build_systems.cmake import CMakePackage
 from spack.package import *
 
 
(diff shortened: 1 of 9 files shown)


ggeorgakoudis:
@cmelone Changes are welcome! Why do we need the weak symbol for static linking? Does Mneme redefine it?

I'm all for upstreaming the package to the Spack repo. How does that work for maintaining it?

packaging/spack/spack_repo/proteus/packages/proteus/package.py
cmelone: is this still a known issue? I was able to use build and use proteus with BUILD_SHARED=ON with cuda support

cmelone:
like I mentioned in #495, I can upstream package.py to spack/spack-packages and turn this PR into removing the package from this repo

packaging/spack/spack_repo/proteus/packages/proteus/package.py
ggeorgakoudis: No, this is resolved and proteus works fine with cuda as a shared library

ggeorgakoudis:
> like I mentioned in #495, I can upstream package.py to spack/spack-packages and turn this PR into removing the package from this repo

Sounds good to me. We need to add the latest release in the description. Please coordinate with @ZwFink.

ZwFink (approved):

ZwFink:
We have decided to merge this as-is so we have updated spack infrastructure. Following, we will upstream the package.py and decide to remove the proteus package.py

PR #503, opened 2026-07-06
Fixup: Ci/matrix integration testing

ZwFink:
We shouldn't rely on the LLVM version available on Matrix, as the LLVM 20 we rely upon was changed for LLVM 21, and LLVM 20 was removed. This causes CI failures in #496 , #467 for example. 

This change installs LLVM 20 in a new miniforge instance during CI.


johnbowen42 (approved):
thanks Zane

PR #518, opened 2026-09-08
Consolidate code generation under Dispatcher

ZwFink:
Moves the annotation-based runtime to use the Dispatcher, which is now shared by all frontends. Also performs various cleanups in the api (introduce `KernelName`, replace `CompileOptions` with explicit hooks).
A side effect of this is that proteus no longer double-optimizes when using HIP RTC paths. This makes small differences Adam's numerics. 

This branch also normalizes `cache-stats` printing behavior between the three frontends, hence the changes to many different test files. The previous behavior was basically incorrect, and cache stats would be printed unconditionally for the MLIR frontend.


src/include/proteus/impl/JitEngineDevice.h
```diff
@@ -641,21 +632,31 @@ JitEngineDevice<ImplT>::compileAndRun(
       LambdaJitValuesMap.empty() ? EmptyLambdaCallsiteRuntimeConstants
                                  : LambdaJitValuesMap;
 
+  auto CreateTask = [&]() {
```
johnbowen42, line 635: the `CreateTask` lambda is great here, I almost wonder if we want to just make it a first class factory function and make it the only way to construct a `CompilationTask` (constructing such a complex class with brace initializers feels sketchy)

src/include/proteus/impl/JitEngineDevice.h
```diff
@@ -669,31 +670,18 @@ JitEngineDevice<ImplT>::compileAndRun(
...
-  CodeCache.insert(HashValue, KernelFunc, KernelInfo.getName());
-  if (CacheChain)
-    CacheChain->store(HashValue,
-                      CacheEntry::staticObject(ObjBuf->getMemBufferRef()));
+  CompiledLibrary Library{std::move(ObjBuf)};
+  Library.GlobalsRelinked = true;
 
-  return launchKernelFunction(KernelFunc, GridDim, BlockDim, KernelArgs,
-                              ShmemSize, Stream);
+  return Launch(LoadKernel(Library));
```
johnbowen42: nit: `launchKernelFunction` was a first-class function before, why not keep and we can drop a lambda?  This way we just have one lambda here, the `LoadKernel` library.  I think three lambdas is a bit much and hurts readability slightly while not really serving a goal requiring lambdas

include/proteus/Frontend/Dispatcher.h
```diff
@@ -75,21 +79,39 @@ struct DispatchResult;
...
   std::unique_ptr<ObjectCacheChain> ObjectCache;
 
   Dispatcher(const std::string &Name, TargetModelType TM);
 
+  void printObjectCacheStats();
+
 public:
   static Dispatcher &getDispatcher(TargetModelType TargetModel);
-  virtual ~Dispatcher() = default;
+  virtual ~Dispatcher();
```
johnbowen42, line 91: Why do we want this destructor to no longer be default?

src/include/proteus/impl/Frontend/DispatcherHIP.h
```diff
@@ -117,90 +112,13 @@ class DispatcherHIP : public Dispatcher {
...
                                            : "oclc_wavefrontsize64_on.bc");
 
-    llvm::Linker Linker{*ModOwner};
+    llvm::Linker Linker{M};
     for (const auto &Path : LibsToLink) {
       auto LibMod = LoadBitcode(Path);
       Linker.linkInModule(std::move(LibMod),
                           llvm::Linker::Flags::LinkOnlyNeeded);
     }
-
```
johnbowen42, line 126: I like that this simplifies the dispatchers

johnbowen42:
Could we do some performance studies about overhead before we merge this, since its on potentially a hot path?  I have accidentally hurt warm-cache performance before modifying this specific code path inside `JitEngineDevice` in particular Not sure what a good proxy would be

johnbowen42 (changes requested):

ZwFink:
> Could we do some performance studies about overhead before we merge this, since its on potentially a hot path? I have accidentally hurt warm-cache performance before modifying this specific code path inside `JitEngineDevice` in particular Not sure what a good proxy would be

In a microbenchmark where we launch a trivial kernel thousands of times, this branch added ~0.3us overhead to JIT launches. The overhead came from building an additional `KernelName` and re-hashing the suffix before cache lookups. After [9509f62](https://github.com/Olympus-HPC/proteus/pull/518/commits/9509f62435b70da5230c699314c691c523c67ff6), performance is identical.

include/proteus/Frontend/Dispatcher.h
ZwFink: Because we have a forward-declared `ObjectCacheChain` in `Dispatcher.h` now. If we include the `default` definition of `~Dispatcher` here, then includers of `Dispatcher.h` will fail when the default `~Dispatcher` tries to destruct an incomplete type (`std::unique_ptr<ObjectCacheChain>`). So, we keep the default implementation of `~Dispatcher` but move it to `Dispatcher.cpp`.

src/include/proteus/impl/JitEngineDevice.h
ZwFink: The lambda does a nice job of deduplication and avoiding unnecessary construction. If we have a pending compilation, we skip `CompilationTask` construction by not calling the lambda. Alternatively, we could construct `CompilationTask` on demand in both branches, but this duplicates the construction code, even if we construct via a factory or builder.

src/include/proteus/impl/JitEngineDevice.h
ZwFink: `launchKernelFunction` still exists and is called in the Dispatcher. But I agree the lambda here adds little. I've replaced it with direct calls.

src/include/proteus/impl/JitEngineDevice.h
johnbowen42: yeah that is a good point, we would still be passing a long list of parameters to the factory function

src/include/proteus/impl/JitEngineDevice.h
johnbowen42: nice!

johnbowen42:
> > Could we do some performance studies about overhead before we merge this, since its on potentially a hot path? I have accidentally hurt warm-cache performance before modifying this specific code path inside `JitEngineDevice` in particular Not sure what a good proxy would be
> 
> In a microbenchmark where we launch a trivial kernel thousands of times, this branch added ~0.3us overhead to JIT launches. The overhead came from building an additional `KernelName` and re-hashing the suffix before cache lookups. After [9509f62](https://github.com/Olympus-HPC/proteus/pull/518/commits/9509f62435b70da5230c699314c691c523c67ff6), performance is identical.

Cool!  I'm glad you measured and pinned that down

johnbowen42 (approved):
LGTM now that the benchmarking has done and performance restored

PR #520, opened 2026-09-10
bugfix:  Support MFEM/RAJA LLVM IR shapes in LLVM analysis

johnbowen42:
Previously, Proteus's LambdaAnalysis could not successfully trace a lambda like this
```
int i = 10;
a = proteus::register_lambda([=] {...});
b = [i, a] (int) {...};
```

which is an LLVM IR shape that was underlying several uses of Proteus within RAJA-enabled MFEM in production codes.  This commit enhances the LLVM pointer use analysis to be offset-aware, meaning that the analysis keeps track of where the closure pointer lives within it's current work item.  In the case above, this allows the analysis to look two separate GEP / store combination (the one storing the integer, and the one storing the registered closure pointer), and disregard the integer store because it does not pertain to the memory location of the registered closure.

The initial offset for an LLVM pointer is given to the `KernelArgUseVisitor` by the `KernelArgVisitor` through the `getDominatingUse` API.  Each ptr is assigned its own offset in a map internal to the `KernelArgUseVisitor`.  Whenever an item is put onto the worklist, it's offset is added to the map.  e.g. `GEPInst` will add their base pointer with the overall GEP's offset taken away from the workitem's offset.


johnbowen42:
I need to add more tests to this PR before we merge it.  I'm debating between including MFEM/RAJA as an integration test, but will probably just end up adding more unit tests.

ZwFink:
> I need to add more tests to this PR before we merge it. I'm debating between including MFEM/RAJA as an integration test, but will probably just end up adding more unit tests.

For now I prefer unit tests due to CI times that are already long.

src/include/proteus/impl/Config.h
```diff
@@ -222,8 +213,9 @@ class CodeGenerationConfig {
         getEnvOrDefaultBool("PROTEUS_SPECIALIZE_ARGS", true),
         getEnvOrDefaultBool("PROTEUS_SPECIALIZE_LAUNCH_BOUNDS", true),
         getEnvOrDefaultBool("PROTEUS_SPECIALIZE_DIMS", true),
-        getEnvOrDefaultBool("PROTEUS_SPECIALIZE_DIMS_RANGE",
-                            DefaultSpecializeDimsRange),
+        // SpecializeDimsRange is disabled by default due to observed
+        // regressions for HIP and CUDA code generation
+        getEnvOrDefaultBool("PROTEUS_SPECIALIZE_DIMS_RANGE", false),
         getEnvOrDefaultChar("PROTEUS_OPT_LEVEL", '3'),
```
ZwFink: Should this be included in this PR?

src/include/proteus/impl/CompilationTask.h
```diff
@@ -61,7 +61,7 @@ class CompilationTask {
     // Parallel codegen forwards custom pipelines to LTO; HIP RTC invokes
     // optimization internally.
     // TODO: Move optimizeIR inside the codegen routines?
-    if (CGOption == CodegenOption::Serial)
+    if (CGOption != CodegenOption::Parallel)
       optimizeIR(M, DeviceArch, OptConfig);
```
ZwFink, line 65: What is this for?

tests/gpu/CMakeLists.txt
```diff
@@ -264,14 +264,19 @@ CREATE_GPU_TEST(lambda_def lambda_def.cpp)
 CREATE_GPU_TEST(lambda_factory lambda_factory.cpp)
 CREATE_GPU_TEST(lambda_factory_forwarded lambda_factory_forwarded.cpp)
 CREATE_GPU_TEST(lambda_thread_privatizer lambda_thread_privatizer.cpp)
+CREATE_GPU_TEST(lambda_index_first_wrapper lambda_index_first_wrapper.cpp)
 CREATE_GPU_TEST(lambda_phi_analysis lambda_phi_analysis.cpp)
+CREATE_GPU_TEST(lambda_insertvalue_offsets lambda_insertvalue_offsets.cpp)
+# CREATE_GPU_TEST(lambda_provenance_shapes lambda_provenance_shapes.cpp)
+# CREATE_GPU_TEST(lambda_store_order lambda_store_order.cpp)
```
ZwFink: Disabled tests

src/pass/KernelArgVisitor.h
```diff
@@ -253,209 +234,117 @@ class LambdaArgVisitor : public InstVisitor<LambdaArgVisitor> {
...
-      // AnalysisSuccess = false;
-      // DEBUG(Logger::logs("proteus-pass")
-      //     << "Analysis failed at = \n"
-      //     << ASC << "\n");
-    }
+
     WorkList.push_back({Res->DominatingWrite, &ASC});
     // Default is zero so we can safely add it
-    Offset += Res->Offset;
+    Offset -= Res->Offset;
```
ZwFink, line 492: Comment is out-of-date now that we subtract

src/pass/KernelArgVisitor.h
```diff
@@ -253,209 +234,117 @@ class LambdaArgVisitor : public InstVisitor<LambdaArgVisitor> {
...
+    //     GetPointerBaseWithConstantOffset(LI.getPointerOperand(), LoadOff,
+    //     DL);
+    // if (!LoadBase) {
+    //   AnalysisFailed = true;
+    //   return;
+    // }
+    // LoadBase = LoadBase->stripPointerCasts();
+    // auto Res = getDominatingUse(DL, LoadBase, MemoryAnalysisPtrUse);
+    // if (!Res)
+    //   return;
```
ZwFink: Comments

src/pass/KernelArgPtrUseVisitor.h
```diff
@@ -74,127 +100,28 @@ getTrackedPointerLocation(const DataLayout &DL, Value *Ptr) {
...
-  CallerFrame ArgBeforeCB;
   int64_t Offset = 0;
+  // The ValueOffsetMap contains the "live range" of the ptr we're analyzing.
+  // For example, let's say that LambdaInstUseVisitor is handed ptr %0 = alloca
+  // ptr, and LambdaArgVisitor has already identified that the closure starts at
+  // byte
+  // 8.  In this case, ValueOffsetMap[%0] = 8.  If we encounter a store like
+  // store ptr %2, ptr%0, align 8, we don't care, because its written outside
+  // of the range of the closure.
+  DenseMap<Value *, int> ValueOffsetMap;
```
ZwFink: Why int instead of i64?

src/pass/KernelArgPtrUseVisitor.h
```diff
@@ -220,23 +155,54 @@ class LambdaInstUseVisitor : public InstVisitor<LambdaInstUseVisitor> {
...
-    }
+    WorkList.push_back(UseEdge{NextVal, CurVal});
+  }
+
+  void offsetValueMapFailure(Value *V) {
+    AnalysisFailed = true;
+    AnalysisSuccess = false;
+    DEBUG(Logger::logs("proteus-pass")
+          << "    [PTR use analysis]: Analysis failed due to absence of " << *V
+          << " in offset tracking map, this is an internal compiler bug\n");
```
ZwFink, line 236: Is it a proteus bug or Clang/LLVM bug if we reach this point?

ZwFink (changes requested):

src/pass/KernelArgPtrUseVisitor.h
johnbowen42: In my opinion this means that we messed something up by not populating an internal data structure correctly

src/pass/KernelArgPtrUseVisitor.h
johnbowen42: I came up with this in a hurry, that is a good point and I will change it

src/include/proteus/impl/CompilationTask.h
johnbowen42: we do RTC by default for HIP, and `optimizeIR` was not getting called and I as getting some poor performance as a result.  It might be possible to revert this line, but I need to rerun some benchmarks

src/include/proteus/impl/Config.h
johnbowen42: Good point I'll make a separate PR

tests/gpu/CMakeLists.txt
johnbowen42: re-enabled and committed these tests

src/include/proteus/impl/Config.h
johnbowen42: https://github.com/Olympus-HPC/proteus/pull/524

src/pass/KernelArgPtrUseVisitor.h
ZwFink: Makes sense. Then I think this should state the bug is in Proteus

src/include/proteus/impl/CompilationTask.h
ZwFink: IIRC RTC already does O3.
(I just double-checked and RTC indeed does O3).

johnbowen42: I've removed these from this PR and moved them to the performance regression PR, I will review taking it out

ZwFink (approved):

PR #524, opened 2026-09-15
performance: Disable range specialization by default

johnbowen42:
previously, the specialization enabled by `PROTEUS_SPECIALIZE_DIMS_RANGE` was enabled by default.  However, this caused serious performance regressions in MFEM's TMOP kernels, as well as production codes.  This PR leaves the specialization but disables it by default


ZwFink (approved):
Tests pass --> Merge, but I think you should remove the change that calls `optimizeIR` with RTC, as we would be double-optimizing.

Issue #382, opened 2026-01-15
Allow setting block/grid dimensionalities on grid stride loop kernels


koparasy:
Currently proteus allows to set the following options:
```json
	"foo" : {
		"CodeGen" : "serial",
		"LaunchBounds" : false,
		"SpecializeDims" : false,
		"SpecializeDimsRange" : false,
		"OptLevel" : 1,
		"CodeGenOptLevel" : 2,
		"Pipeline" : "default<O3>,globaldce",
		"TunedMaxThreads" : 1,
		"MinBlocksPerSM": 4
	}
```

In the case of grid stride loop kernels tuners can configure optimal GridDim and BlockDim. There is currently no support to forward these values to proteus.

PR #528, opened 2026-09-18
Launch params in PROTEUS_TUNED_KERNELS

Dando18:
Addresses #382 by allowing passing grid/block dims to proteus. Currently, makes them optional members of the config. Allows passing something like below to proteus:

```json
{
  "tuned_dims": {
    "GridDim": { "x": 2, "y": 3, "z": 1 },
    "BlockDim": { "x": 4, "y": 2, "z": 1 },
    "TunedMaxThreads": 8,
    "MinBlocksPerSM": 2
  }
}
```

Since this can potentially change semantics and make a kernel incorrect (e.g. a kernel without grid striding) I'm in favor of adding some other flag to make this a bit more explicit or "opt-in". For example, `PROTEUS_ALLOW_TUNED_LAUNCH_DIMS=true|false`. Happy to add this, just let me know.

This will help ease some of the infrastructure in Mneme/R3. For R3, it'd be nice to also be able to pass IR, but I can save that for future thought.


ZwFink:
I'm not sure we need to do more to make opt-in. Isn't it already opt-in because you have to specify it in a config file?

Dando18:
changing grid/block parameters could lead to correctness issues, so it could be good to have a separate path for opting-in to them.

But I think it's fine either way.

